PR #5, opened 2019-09-30
Add ZEN waypoint config

jokoho48:
- Add ZEN waypoint config
- cleanup wp functions


nk3nny:
- Zen waypoint support? Fantastic. 
- fn_taskCamp is not currently in use. It is however part of my regular mission making framework, so hence been merged for future use and reference.

PR #6, opened 2019-10-01
Add Debug Renderer

jokoho48:
The version in the video still had some bugs, but they should be fixed now
Preview video: https://youtu.be/2U6fLC8NWxk


nk3nny:
Cool interface.  It would be a very useful expansion to mark group leaders in some manner and perhaps have a way to display information stored in the dangerMode variable.  (DangerMode queue and timeouts mainly). Either through this interface, or a similar one.

jokoho48:
yes will do that.

jokoho48:
is the DangerMode Only Important for the Leader or does every AI needs that?

jokoho48:
i now added the DangerMode Stuff to Only Squad Leaders

jokoho48:
https://i.imgur.com/XTC5FKY.jpg

nk3nny:
_DangerMode_ is relevant mostly for the leader.  First array contains information relating to what the Squad has assessed earlier. 

Fourth entry gives information when the DangerMode will time out. That is-- the group will respond to contact by 'reacting' again. 

Of these, the first one is the most relevant

nk3nny:
First entry is an array which is filled with numbers.  There will only be one entry of each number. Order of priority is lower is better. 

1, "Contact"
2, "Hide from tank/air craft"
3, "Manoeuvre" 
5, "Check nearby buildings"
6, "Call artillery"

At the current time. The intention is to add to these to expand more customized behavior. Such as  'Black Hawk Down' esque  swarms.

nk3nny:
Picture is perfect.  Though for the sake of clarity, may be better to mark the unit as **group leader** rather than **squad leader** (which carries different connotations in the context of Arma.

nk3nny:
Will happily Merge once ready.

PR #12, opened 2019-10-01
Add function "findCover"

diwako:
Adds the function `lambs_danger_fnc_findCover`.
This functions checks close to the unit a spot it potentially seek cover from a given enemy.
The function seeks the closest possible location.

As return value an array is returned which consists of the Position in AGL and as second item the stance. Combined it should be possible to a unit to seek cover.

Right now this function is not implemented into the FSM, yet it is there in case a more dedicated method of finding cover is ever needed.

### DIFF
--- a/addons/danger/XEH_PREP.hpp
+++ b/addons/danger/XEH_PREP.hpp
@@ -31,3 +31,4 @@ PREP(vehicleAssault);
 PREP(vehicleJink);
 PREP(vehicleRotate);
 PREP(vehicleSuppress);
+PREP(findCover);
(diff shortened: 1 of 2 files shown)


nk3nny:
Cool function. Would it be possible to create a vehicleFindCover adaption which would allow tracked vehicles-- tanks essentially-- to find hulldown positions? 

Searching for walls, buildings and dips in the terrain. Preferably allowing for the size of the vehicle

nk3nny:
Actually.. Riffing of the same though.  A variant suitable for Attack helicopters but then _only_ looking at topography and setting a flyInHeight feature to allow for peeking

diwako:
This function is to be considered for infantry only right now. It uses terrain objects and editor placed objects as reference in order to find a position.

Adding topography into this would make this a very expensive to run function. For ground vehicles I could see a dedicated function which just checks for large rocks, houses and maybe trees and try to find a position that cannot be seen from multiple angles to the danger position.
Yet adding the possibility to scan for hills to hide behind and ditches to hide in would require checking quite a few more.

For helicopter that would also mean to check tree heights and clearings a chopper can fly low enough to not hit a tree and hide as well. Let alone if the AI would actually allow such maneuvers.

PR #14, opened 2019-10-01
Add CBA Event 'lambs_danger' on danger start

veteran29:
Adds `lambs_danger` CBA local event fired when AI detects danger.

This adds a possibility to have more than one handler in opposition to `dangerCode`.

### DIFF
--- a/addons/danger/scripts/lambs_danger.fsm
+++ b/addons/danger/scripts/lambs_danger.fsm
@@ -1229,8 +1229,10 @@ class FSM
                          "// change formation " \n
                          "group _this setFormation (group _this getVariable [""dangerFormation"",formation _this]);" \n
                          "" \n
-                         "// execute code " \n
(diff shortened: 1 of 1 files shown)


diwako:
Right now we are currently changing a lot of variable names, The `dangerCode` variable will also be adapted at some point. That means we might will ask for different event name, but we'll see.
We are also discussing more CBA events through out all of the states, so people can hook into the AI or read what is going on.

So just as a heads up in case this PR will be just sitting there for a bit.

jokoho48:
#23

diwako:
Just wanted to inform you that a local CBA event will be thrown called `lambs_danger_onContact`

The parameter this event receives will be `[_this, group _this, units _this]`

This is also a bit awkward right now, as this PR is now obsolete...

veteran29:
Lets close it then.

Issue #24, opened 2019-10-04
Off target cannon fire and turreted wheeled transports


nk3nny:
**fn_vehicleAssault.sqf** 
Cannon fire is determined by a highly random formula. One thing that it does not take into account is whether the turret is reasonably aligned-- say with 10 degrees-- of the target in question. 

I appreciate the chaos added by off-target shooting. But completely off track is not ideal. 

**Tank fsm** 
Currently only tracked vehicles are picked out for the vehicleAssault script. This leaves out otherwise powerful vehicles, like BTRs, LAV-25s, or vehicles with similar capability, but no tracks. 

I guess sorting for ~"transported_wheeled" (or whatever the class is), might settle it however.

Issue #9, opened 2019-10-01
Improve danger waypoints


nk3nny:
**Misc improvements**
- [x] Update systemchat debug messages to match format of the rest of the mod
- [x]  Reorganise waypoint scripts. Current wp scripts to become _functions_. Add new wp scripts which instead call these functions.  In this manner the taskX/Y functions may be called from elsewhere too. 

**Extend Waypoint features**
- [x]  ~Add waypoint specific UI in EDEN for relevant options:~ 
~Examples: ~
    * ~Garrison: 'teleport into position'~
    * ~TaskRush/Creep/Hunt to use group leader pos as center instead~
    * ~TaskRush/Creep/Hunt to use side knowsAbout as limit rather than range~
- [x] Tweak some of the default ranges. These tend to be on the smaller side. 

**ZEUS or ZEN implementation**
- [x] ~Add vanilla ZEUS support to current waypoints~ <-- Discarded for version 2.0

----

**Feature additions** 

- [ ] Add Assault Waypoint https://github.com/nk3nny/LambsDanger/issues/21
- [ ] Add Dynamic Base of Fire Waypoint
- [ ] Add Retreat Waypoint

nk3nny:
https://github.com/nk3nny/LambsDanger/issues/21

PR #27, opened 2019-10-05
A series of tactical changes relating to 1.4.5

nk3nny:

1. Tactical changes relating to feedback from Forums, Video and Steam Workshop
2. Practical changes to a number of functions
3. Beginning of waypoint conversions
4. Removes unused functions

- Converts WP to standard related to https://github.com/nk3nny/LambsDanger/issues/9
- As noted adds some infantry movement considerations. Especially in terms of increasing aggression. 
- Added Wheeled APC support
- Added limits to turret fire https://github.com/nk3nny/LambsDanger/issues/24


addons/wp/CfgWaypointTypes.hpp
```diff
@@ -7,37 +7,36 @@ class ZEN_WaypointTypes {
...
         displayName = "5 TASK CREEP";
         type = "SCRIPTED";
-        script = QPATHTOF(functions\fnc_taskCreep.sqf);
+        script = QPATHTOF(scripts\fnc_wpCreep.sqf);
     };
     class EGVAR(danger,CQB) {
         displayName = "6 TASK CQB [Experimental]";
         type = "SCRIPTED";
-        script = QPATHTOF(functions\fnc_taskCQB.sqf);
+        script = QPATHTOF(scripts\fnc_wpCQB.sqf);
```
diwako: This file is currently missing and would lead to an error.

addons/wp/CfgWaypoints.hpp
```diff
@@ -11,42 +11,42 @@ class CfgWaypoints {
         class EGVAR(danger,Garrison) {
             displayName = "1 TASK GARRISON";
             displayNameDebug = QEGVAR(danger,Garrison);
-            file = QPATHTOF(functions\fnc_taskGarrison.sqf);
+            file = QPATHTOF(scripts\fnc_wpGarrison.sqf);
```
diwako: Missing file

addons/wp/CfgWaypoints.hpp
```diff
@@ -11,42 +11,42 @@ class CfgWaypoints {
...
             icon = "\a3\3DEN\Data\CfgWaypoints\Guard_ca.paa";
             tooltip = "Garrisons units in buildings around waypoint out to WP completion radius (default 50m)";
         };
         class EGVAR(danger,Patrol) {
             displayName = "2 TASK PATROL";
             displayNameDebug = QEGVAR(danger,Patrol);
-            file = QPATHTOF(functions\fnc_taskPatrol.sqf);
+            file = QPATHTOF(scripts\fnc_wpPatrol.sqf);
```
diwako: Missing file

addons/wp/CfgWaypoints.hpp
```diff
@@ -11,42 +11,42 @@ class CfgWaypoints {
...
+            file = QPATHTOF(scripts\fnc_wpRush.sqf);
             icon = "\a3\3DEN\Data\CfgWaypoints\Sentry_ca.paa";
-            tooltip = "Rushes towards enemy with WP completion radius (Default 500m)";
+            tooltip = "Rushes towards enemy with WP completion radius (Default 1000m)";
         };
         class EGVAR(danger,Hunt) {
             displayName = "4 TASK HUNT";
             displayNameDebug = QEGVAR(danger,Hunt);
-            file = QPATHTOF(functions\fnc_taskHunt.sqf);
+            file = QPATHTOF(scripts\fnc_wpHunt.sqf);
```
diwako: Missing file

addons/wp/CfgWaypoints.hpp
```diff
@@ -11,42 +11,42 @@ class CfgWaypoints {
...
             icon = "\a3\3DEN\Data\CfgWaypoints\Sentry_ca.paa";
-            tooltip = "Hunts the AI around group leader out to WP completion radius (Default 500m)";
+            tooltip = "Hunts the AI around group leader out to WP completion radius (Default 1000m)";
         };
         class EGVAR(danger,Creep) {
             displayName = "5 TASK CREEP";
             displayNameDebug = QEGVAR(danger,Creep);
-            file = QPATHTOF(functions\fnc_taskCreep.sqf);
+            file = QPATHTOF(scripts\fnc_wpCreep.sqf);
```
diwako: Missing file

addons/wp/CfgWaypointTypes.hpp
diwako: Also do you really want to move these to scripts? having them compiled as functions would be better, so mission makers can just use them in scripts.

nk3nny: 1. Missing files may be due to user error on my part. Perhaps I didn't mark everything when pushing on SmartGit. (Not smart). Unfortunate.  I'll have that sorted when I get off work. 

2. In regards to the fnc_wpCQB.sqf.  The _fnc_ moniker may be wrongly chosen.  Strictly speaking I do not see any utility in keeping these as functions. Their only purpose is to exist within waypoints-- which are essentially execVMs as I understand it. 

If scripters or module makers need access to the tactical functions, it would be better to call these directly as functions. The _waypoint functions_ exist only as staging point to call or spawn the functions as appropriate function.

addons/wp/CfgWaypoints.hpp
nk3nny: A brief description of the fnc_wpXxx.sqf  files

These adapt information from the waypoint: group, position and radius,  as well as any future WP specific settings and variables-- on the model of the 'advanced mine sweeping' wp in vanilla,  and parse them into the relevant script. 

1. This divide means that the taskXxx function can remain unencumbered by any WP specific checks. It is a function which can be called anywhere without any connection to a specific waypoint.  As noted above. This adds utility and makes maintaining the function in a 'universal' manner much easier. 

2. Noted exception is the 'Register Artillery' wp-- which really only does one thing.  It might be moved into script, but it is conceivably something that can be called as a function by MM.  

3. There is nothing wrong with moving all /scripts into /functions and compiling them. I just did not see the need for three reasons: 
3.1 - Intended use outlined above. 
3.2- Easier to keep distinct in the design process. 
3.3- Not necessary to keep constantly in memory. 

-k

addons/wp/CfgWaypoints.hpp
nk3nny: fixed

addons/wp/CfgWaypoints.hpp
nk3nny: fixed

addons/wp/CfgWaypoints.hpp
nk3nny: fixed

addons/wp/CfgWaypoints.hpp
nk3nny: fixed

Issue #23, opened 2019-10-04
Implement Event System


jokoho48:
I personally thing an Event system is an important thing, but I am not sure if the CBA Event system is really fitting for this purpose. I would rather have a Mix out of the Vanilla Event System and the CBA one.
a Calling Example would be

FSM:  
This Spawn in this case I needed so that we don't overload the FSM Scheduler time and then we switch back to Unscheduled
```sqf
[_this, group _this, units _this] spawn {
    [{
         ["lambs_danger_OnDanger", _this] call CBA_fnc_localEvent;
    }, _this] call CBA_fnc_directCall;
};
```

UserEnd:
```sqf
["lambs_danger_OnDanger", {
    hint "Awsome Event AI Unit Things " + str _this;
}] call CBA_fnc_addEventHandler;
```

Possible Events:
- OnDanger/OnContact
- OnArtilleryCalled
- OnCQBEntered
- OnPanic
- OnInformationShared
- OnCheckBody

Edited/Update to design we want to use

veteran29:
I also don't see a need to support both event systems at once, this is redundant.

nk3nny:
I suspect anyone savvy enough to leverage eventhandlers will be capable of using the CBA ones. I'm not quite seeing the advantage of the Bohemia one? ELI5

jokoho48:
closed and implemented as of #28

PR #28, opened 2019-10-06
Event Callback for Actions

jokoho48:

Add Event Callbacks for Actions

- [x] OnDanger/OnContact
- [x] OnArtilleryCalled
- [ ] OnCQBEntered
- [x] OnPanic
- [x] OnInformationShared
- [x] OnCheckBody

I am currently not sure where to best call OnCQBEntered. Because we have Multiple possible Entry points for that

#23


addons/danger/scripts/lambs_danger.fsm
```diff
@@ -1232,10 +1232,10 @@ class FSM
                          "[_this,_target] call lambs_danger_fnc_shareInformation; " \n
                          "" \n
                          "// change formation " \n
-                         "group _this setFormation (group _this getVariable [""lambs_danger_dangerFormation"",formation _this]);" \n
+                         "(group _this) setFormation (group _this getVariable [""lambs_danger_dangerFormation"",formation _this]);" \n
                          "" \n
-                         "// execute code " \n
-                         "[_this,units group _this] call ((group _this) getVariable [""lambs_danger_dangerCode"", {}]);" \n
+                         "// call event system" \n
+                         "[""lambs_danger_onContact"", [_this, group _this, units _this]] call lambs_danger_fnc_eventCallback;" \n
```
diwako: @nk3nny do we need re-assessment to remove the call for `dangerCode` or can it be directly removed?

jokoho48: I did not found any reference  to the variable in code

nk3nny: Hmm. Every tweak of the FSM adds an odd overhead to the process. I suspect this FSM is different from the one I run in my pull request? One line of code is in any case no big thing. 

In any case. "dangerCode" was _only_ referenced on that point in the danger.fsm. It is only run when the group leader formally calls 'contact'

REACT (individuals hide) --> CONTACT (Group leader) --> ASSESS (Group leader) --> MISC TACTICAL ACTIONS (entire group)

There is a delay between REACT and CONTACT based on the squad leaders individual situation. 
There is a delay between CONTACT and ASSESS based on groups timing scheme-- 30-60 seconds 
There is delay between ASSESS and TACTICAL ACTIONS based on squad leaders timing scheme

If there are no potential TACTICAL ACTIONS  (as found by group leader ASSESS).  Then the group leader will continue to attempt periodic ASSESS. 

CONTACT is only called on 'first contact' or after 6 minutes without danger mode active.

jokoho48:
we can always add more events later when we see a point or find where it would fit in.

PR #29, opened 2019-10-06
add Improved FindOverwatch Position

jokoho48:

1. *Adds an Improved FindOverWatch Position Function*

[FindOverwatchTestMission.Altis.zip](https://github.com/nk3nny/LambsDanger/files/3697330/FindOverwatchDemo.Altis.zip)


nk3nny:
I like this. Can we can a comparison of the types of positions found by this function in comparison to the vanilla one?

PR #30, opened 2019-10-06
Introduce Sort Mode parameter for find cover

diwako:

Introduce a new parameter to the `findCover` function.

The new parameter Accepts only a string with possible values:

- ASCEND `default`
- DESCEND 
- RANDOM

If neither `DESCEND` or `ASCEND` is provided then the possible positions are being randomized. `ASCEND` returns closest possible location. `DESCEND` returns farthest possible location.

### DIFF
--- a/addons/danger/functions/fnc_findCover.sqf
+++ b/addons/danger/functions/fnc_findCover.sqf
@@ -8,6 +8,7 @@
  * 1: Enemy <OBJECT> or Enemy Position (AGL) <ARRAY>
  * 2: Range to find cover, default 15 <NUMBER>
  * 3: Sort mode <STRING>, default "ASCEND", possible values: ASCEND, DESCEND, RANDOM. ASCEND returns closest possible location
+ * 4: Max Results <Number>, default 1, Maximum amount of results that can be returned, -1 for all (warning may be slow)
  *
  * Return Value:
  * Array of format [_posAGL, _stance], when no cover found then an empty array is returned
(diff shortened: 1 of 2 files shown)


nk3nny:
Would it be possible to get a variable to return a number of entries? Humans don't always make perfect choices and neither should the AI. 

Essentially, the ability to get  the 'three nearest' or 'six random spots' or 'one furthest'. 

-k

nk3nny:
BTW: in the latest FSM pushed in my commit this function is in use in individual the 'coward' behaviour as a test case.

diwako:
That is possible. The question is however what will the return value look like. Right now it looks like this
```sqf
[[0,0,0],"UP"]
```
If nothing is found it is just an empty array, which will stay.

I can imagine this will look like this as well if you only want one position, but what about multiple?
Something like this?
```sqf
[
    [[0,0,0],"UP"],
    [[0,0,0],"UP"],
]
```

The reason is, I would like to have all return value patterns the same. That would mean however you need to get the first value out of it if you just want the first position

nk3nny:
Second options seems natural to me. Just strikes me that as another advantage of this function-- it can be called on the group and thus have them scatter around quite naturally. (with no two occupying the same position). 

To be honest, for most application I think the stance is unecessary-- something which (because the fsm generally calls 'setUnitPosWeak' will be overruled anyway.  A flag to skip the check might be a tiny performance enhancement. 

That said. I love the function. This may very well be the basis of a formationEntity.fsm replacement!

diwako:
The reason why it gives back a stance is as it checks from the possible cover position from feet up to head height. Reason for this is, it might consider Street signs a cover position while only the head is in cover other wise. So it gives a usable stance as well, to make sure that unit can be at least concealed.

nk3nny:
Good call!

diwako:
Pushed changes, here is what it would return now.
```sqf
[
    [[1724.04,1037.54,0.259367],"MIDDLE"],
    [[1732.6,1036.6,0.259367],"MIDDLE"],
    [[1744.11,1044.82,0.166372],"UP"],
    [[1739.7,1047.69,0.166372],"UP"],
    [[1736.01,1046.54,0.166372],"UP"],
    [[1738.14,1020.18,0],"DOWN"],
    [[1732.56,1018.42,0],"MIDDLE"],
    [[1732.87,1023.65,0],"UP"]
]
```

PR #40, opened 2019-10-12
Do some Cleanup

jokoho48:

1. Do Some Cleanup
2. Fix an issue with Buttons
3. Improve ShareInformationRange


addons/danger/XEH_preInitClient.sqf
```diff
@@ -36,11 +36,11 @@ _suppress_AI = {
 };
 
 // functions ~ easy hide
-_hide_AI = {
+private _fnc_hide_AI = {
     private _buildings = [player,38,true,true] call FUNC(findBuildings);
     private _units = (units player) select {_x distance player < 55 && {!isPlayer _x}};
     {
-        [_x,_pos,_range,_buildings] call FUNC(hideInside);
+        [_x, getPos _x, 45, _buildings] call FUNC(hideInside); // What should be the Default Distance here? @nkenny?
```
nk3nny: Just set it to _nil_ or 10 meters. The default range needs revisiting in any case.

addons/danger/functions/fnc_shareInformationRange.sqf
```diff
@@ -30,27 +30,22 @@ _range = switch (side _unit) do {
 _range = _range min viewDistance;
 
 // Sort long range radios
-_radio = false;
-{
-    // in vehicle (and close), has backpack, or TFAR radio
-    // or lambs_danger_fnc_dangerRadio is set to true
-    _hasRadio = (
-        _x getVariable [QGVAR(dangerRadio),false]
+private _target = objNull;
```
nk3nny: Am I misreading something here? If the function finds no unit with a backpack, it still needs to return the calling unit. That unit will be the origin of the radio transmission.

From a design point of view: The backpack/variable function exists just on the border of the scope. Adding a backpack is a mission makers option, but not integral to the functioning of the mod. Hence the generous default communications range.

The reason this function needs to return Unit, range and radio, is to allow for scripts that will only trigger when a long range radio is present. (Which might be useful in an artillery/support rewrite).

jokoho48: aha now it makes more sense what that function does do. will be fixed in next push

addons/danger/XEH_preInitClient.sqf
jokoho48: 3 lines about you checked in a distance of  38 units. and I added a bit more that is why I did use 45. but yeah 10 it then is

addons/danger/functions/fnc_shareInformationRange.sqf
nk3nny: #needs_better_code_headers

addons/danger/XEH_preInitClient.sqf
nk3nny: Will probably need adjustment based to fit game play concerns.  It could have been a CBA setting-- but I don't want the bloat. And we're not a UI mod anyway.

PR #43, opened 2019-10-16
Fix file headers and perform minor tweaks

nk3nny:
- Added new file headers to all files
- Fixed uncommented line in settings.sqf
- Tweaked hide inside range changed from 50 to 3
- Removed automatic suppression for player led groups


jokoho48:
i would set all functions in Danger to Private and not as a Public API. but put all WP function to public. and also the scripts in WP

nk3nny:
wilco

nk3nny:
Suggest we merge this. Will do so within 3hrs unless any major complaints.

PR #56, opened 2020-01-02
Add CQB Formations as settings

jokoho48:

1. *Title*

One thing we need to think about is if we want to Move the CQB Formation settings into an extra Main Cat or not?
@nk3nny, we may also want a more Descriptive Tooltip for the settings!


nk3nny:
Proposed Description: 

"Configure CQB formations"

"Units set to CQB formations will methodically clear buildings when an enemy is encountered"

PR #62, opened 2020-01-20
Tweaks to hemtt

nk3nny:
- Version number 2.0.3
- Possible fix to fn_react to prevent players going prone

2. *Each change in a seperate line*

### DIFF
--- a/.hemtt/base.toml
+++ b/.hemtt/base.toml
@@ -1,0 +1,9 @@
+name = "lambs_danger"
+prefix = "lambs_danger"
+author = "LAMBS Dev Team"
+template = "yes"
+mainprefix = "z"
+include = ["./include"]
+files = ["mod.cpp"]
+folder_optionals = true
+sig_version = 3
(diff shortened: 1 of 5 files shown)


nk3nny:
This is close to the current version 2.0.2 on steam.  I suggest it be discarded. Adding only the tweaks to fn_react. 

I see now that I must have been behind one step on the hemtt.tmol for some reason.  Leaving it here for reference. 

-k

.hemtt/base.toml
```diff
@@ -0,0 +1,9 @@
+name = "lambs_danger"
+prefix = "lambs_danger"
```
diwako: Wouldn't that change the prefix for everything, so addon main is then `lambs_danger_main` and the addon danger is then `lambs_danger_danger` ?

jokoho48: yes I would not change that. also the "new" Template is as far I understood beeing reworked so i would not push that currently

jokoho48:
i would Discard all changes expect the sqf file and use current master for continued dev

PR #64, opened 2020-01-26
Possible fix for Player Moving about without there own input.

jokoho48:
This is a Very Sledge-hammer method of fixing this issue and this needs heavy testing and maybe even removing of some isPlayer checks


nk3nny:
I won't be merging this initially. I'll push out 2.0.3. If the problem persists. I will have another look. 

-k

PR #71, opened 2020-02-19
Optimize fnc_findClosedTarget

dedmen:

1. Make fnc_findClosedTarget faster

### DIFF
--- a/addons/wp/XEH_PREP.hpp
+++ b/addons/wp/XEH_PREP.hpp
@@ -1,4 +1,4 @@
-PREP(findClosedTarget);
+PREP(findClosestTarget);
 PREP(taskArtillery);
 PREP(taskArtilleryRegister);
 PREP(taskAssault);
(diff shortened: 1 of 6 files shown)


dedmen:
Also isn't the function name wrong?? A "Closed" target? That's not what this function does, the description is also wrong.

jokoho48:
good catch @dedmen ... that goes properly on my head. feel free to fix the naming and header.

dedmen:
its a public function though, so renaming might cause issues?

jokoho48:
I think that was wrongly marked as public. we can just mark as private and change the name :P

nk3nny:
Most of the functions in the danger_wp pbo are due a rewrite and rethink to match end user expectations. Wrongly marked public indeed. 

I will hold off merging until we've had a meeting in regards to direction and overall structure. Changes look good however.

jokoho48:
@nk3nny I still think we should merge it because this will be a function that does not really will change. it is a simple getter/find function

nk3nny:
You are absolutely right, @jokoho48. 

-k

PR #73, opened 2020-02-19
Optimize fnc_findBuildings

dedmen:

1. Very slightly make it faster

count makes no sense if you need to add a true/false, its slower than forEach that way.
Also in other scripts the semicolon is usually left away on a return.


jokoho48:
lgtm

PR #74, opened 2020-02-21
add Setting to Disable Gestures

jokoho48:

1. *Title*
@nk3nny please check what playMoveNow can be changed to the gesture function and what functions need to be forced.
https://github.com/nk3nny/LambsDanger/pull/74/files#diff-9c21ba0c96b18cb1cdad7fd182e4fcfbR90
https://github.com/nk3nny/LambsDanger/pull/74/files#diff-aa2356a28d779ef1d576b7aabf6f4918R53


nk3nny:
Wilco. 

Are we adding this setting just for the sake of adding another setting. Or is this something actually requested?  I don't mind, but also don't want settings bloat :P

PR #83, opened 2020-03-16
2.3.1 Hotfixes

jokoho48:

1. *Title*

### DIFF
--- a/addons/danger/functions/fnc_assaultBuilding.sqf
+++ b/addons/danger/functions/fnc_assaultBuilding.sqf
@@ -69,8 +69,12 @@ private _building = (_buildings select 0);
 
 // find spots
 private _buildingPos = _building getVariable [QGVAR(CQB_cleared_) + str (side _unit), (_building buildingPos -1) select {lineIntersects [AGLToASL _x, (AGLToASL _x) vectorAdd [0, 0, 4]]}];
+
 private _buildingPosSelected = _buildingPos select 0;
 
+if (isNil "_buildingPosSelected") then {
+    _buildingPosSelected = _building modelToWorld [0,0,0];
+};
 // remove current target and do move
 _unit lookAt AGLtoASL _buildingPosSelected;
 _unit doMove (_buildingPosSelected vectorAdd [0.5 - random 1, 0.5 - random 1, 0]);
--- a/addons/danger/functions/fnc_leaderArtillery.sqf
+++ b/addons/danger/functions/fnc_leaderArtillery.sqf
@@ -33,15 +33,13 @@ _artillery = _artillery select {
 };
 
 // exit on no ready artillery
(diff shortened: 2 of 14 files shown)


addons/danger/functions/fnc_assaultBuilding.sqf
```diff
@@ -69,8 +69,12 @@ private _building = (_buildings select 0);
 
 // find spots
 private _buildingPos = _building getVariable [QGVAR(CQB_cleared_) + str (side _unit), (_building buildingPos -1) select {lineIntersects [AGLToASL _x, (AGLToASL _x) vectorAdd [0, 0, 4]]}];
+
 private _buildingPosSelected = _buildingPos select 0;
 
+if (isNil "_buildingPosSelected") then {
```
nk3nny: We can keep this in, but should never be a problem  as the findBuilding function (which assigns buildings in the first place in fn_leaderCQC.sqf) will only sort buildings with valid positions. 

I guess theoretically two units could try to remove the same position at the same time however. *shrug*

addons/danger/functions/fnc_leaderArtillery.sqf
```diff
@@ -33,16 +33,14 @@ _artillery = _artillery select {
 };
 
 // exit on no ready artillery
-if (_artillery isEqualTo []) exitWith {if (GVAR(debug_functions)) then {format ["%1 Artillery failed -- no available artillery", side _unit] call FUNC(debugLog);}};
+if (_artillery isEqualTo []) exitWith {if (GVAR(debug_functions)) then {format ["%1 Artillery failed -- no available artillery in rage of Target", side _unit] call FUNC(debugLog);}};
```
nk3nny: "no available artillery in raNge of Target"

addons/danger/functions/fnc_assaultBuilding.sqf
jokoho48: if the variable is nil and the default is an Empty Array because it got filtered out we have to catch that.

addons/danger/functions/fnc_suppress.sqf
```diff
@@ -36,7 +36,7 @@ if (!_override) then {
 
     _pos = AGLtoASL (_unit getHideFrom (_unit findNearestEnemy _unit));
     private _vis = lineIntersectsSurfaces [eyePos _unit, _pos, _unit, vehicle _unit, true, 2];
-    if (count _vis > 1) then {_pos = (_vis select 0) select 0;};
+    if !(_vis isEqualTo []) then {_pos = (_vis select 0) select 0;};
```
nk3nny: Should final argument in line 38 also be 1 instead of two then.  No need to return two positions. ATM I cannot recall why I set it up in that manner.

addons/danger/functions/fnc_leaderArtillery.sqf
jokoho48: i just copied it from the one blow

PR #96, opened 2020-03-24
Add Suppression Check for Friendly Units

### LINKED ISSUE #93: Suppress function should be checking are friendly units nearby
Suppress function is not checking that suppression target is near friendly units. It may cause friendly fire. The function should check is any friendly unit in a proximity ~20 meters of the target to avoid friendly fire. One check at beginning should be enough to fix this issue.

jokoho48:

1. *Title*
2. *Fixes #93*
3. *Replaces #94 and #95*


nk3nny:
Fixed stringtables and altered default settings. 
Reduced 'max abstract range' to from 350 to 280 in suppression-- Reason: More weapons will engage with more aggression at that range

PR #97, opened 2020-03-27
add Player Only Parameter to Creep, Hunt and Rush

jokoho48:

1. *Title*

### DIFF
--- a/addons/wp/functions/fnc_findClosestTarget.sqf
+++ b/addons/wp/functions/fnc_findClosestTarget.sqf
@@ -16,7 +16,7 @@
  * Public: No
 */
 
-params ["_group", ["_radius", 500], ["_area", [], [[]]], ["_pos", [], [[]]]];
+params ["_group", ["_radius", 500], ["_area", [], [[]]], ["_pos", [], [[]]], ["_onlyPlayers", true]];
 
 private _groupLeader = leader _group;
 private _sideExclusion = [side _group, civilian, sideUnknown, sideEmpty, sideLogic];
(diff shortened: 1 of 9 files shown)


addons/wp/stringtable.xml
```diff
@@ -172,6 +172,12 @@
         <Key ID="STR_Lambs_WP_Module_TaskCreep_MovingCenter_ToolTip">
             <English>The creeping group will use the group leader as a center for the search pattern. Disable to have the unit use the module position instead. This creates a more defensive group</English>
         </Key>
+        <Key ID="STR_Lambs_WP_Module_TaskCreep_PlayersOnly_DisplayName">
+            <English>Players Only</English>
+        </Key>
+        <Key ID="STR_Lambs_WP_Module_TaskCreep_PlayersOnly_ToolTip">
+            <English>The creeping group will only Use Players as targets.</English>
+        </Key>
```
diwako: Why not just one translation for this? Means we have for the none tool tip duplicates. Might as well add placeholders and for `format` on the translations

addons/wp/functions/fnc_findClosestTarget.sqf
```diff
@@ -16,27 +16,28 @@
  * Public: No
 */
 
-params ["_group", ["_radius", 500], ["_area", [], [[]]], ["_pos", [], [[]]]];
+params ["_group", ["_radius", 500], ["_area", [], [[]]], ["_pos", [], [[]]], ["_onlyPlayers", true]];
```
diwako: Would set the default value of `_onlyPlayers` to false as it affects more than the modules

jokoho48: No, before Only player were used. now allUnits can be used. to not break backwards compatibility it needs to be true

addons/wp/stringtable.xml
jokoho48: I wanted to keep it the same with all the other modules and later have a pass to remove all duplicates.

PR #110, opened 2020-04-15
Added various performance optimisations to FSM

nk3nny:
Added various performance optimisations to FSM
Added 'Immediate Action' state
Added smarter routines to immediate action
Added fnc_indoor to only recognise buildings
Added units with ACE3 medical AI enabled will exit FSM

Fixed assaulting units from occasionally looking into the celing
Fixed global variable in FSM (timeout)
Fixed Increased APC dismount range
Fixed tightened leaderCQB range (No need to check out to 250 meters)

Changed moved various settings such as speed and search range moved from FSM to function for easier editing
Changed threshold for do suppression while suppressed from 0.5 to 0.75
Changed fsm priorities for bulletClose increased from 3 to 6
Changed fnc_fsmExitVariables to use fnc_isAlive

Removed 'lambs_danger_isChecked' variable from fnc_checkBody (experiments have shown this variable to be unecessary)
Removed 'local' and 'isplayer' check on dangerFSM (unecessary)


addons/danger/functions/fnc_immediateAction.sqf
```diff
@@ -5,77 +5,87 @@
...
+
+// dodge
+_unit setVariable [QGVAR(currentTask), "Dodge!"];
 
-// prone -- exit quickly
+// prone override
 if (_stance isEqualTo "PRONE") exitWith {
-    [_unit, ["EvasiveLeft", "EvasiveRight"], true] call FUNC(gesture);
-    true
+    _unit playActionNow (["EvasiveLeft", "EvasiveRight"] select (_dir > 330));
```
jokoho48, line 29:
```suggestion
    [_unit, [["EvasiveLeft"], ["EvasiveRight"]] select (_dir > 330), true] call FUNC(gesture);	
```

addons/danger/scripts/lambs_danger.fsm
```diff
@@ -498,7 +500,9 @@ class FSM
                                         priority = 5.000000;
                                         to="Artillery__none";
                                         precondition = /*%FSM<CONDPRECONDITION""">*/""/*%FSM</CONDPRECONDITION""">*/;
-                                        condition=/*%FSM<CONDITION""">*/"_artillery"/*%FSM</CONDITION""">*/;
+                                        condition=/*%FSM<CONDITION""">*/"_commander" \n
+                                         "&& {_vehicle getVariable [""lambs_danger_isArtillery"", getNumber (configFile >> ""CfgVehicles"" >> (typeOf _vehicle) >> ""artilleryScanner"") > 0]}" \n
```
jokoho48, line 504: if we do this we maybe want to always set the variable in taskArtilleryRegister?

addons/danger/functions/fnc_immediateAction.sqf
```diff
@@ -5,77 +5,87 @@
...
 if (
-    !(_stance isEqualTo "STAND")
-    || {GVAR(disableAIImediateAction)}
-    || {primaryWeapon _unit isEqualTo ""}
-    || {!(primaryWeapon _unit isEqualTo currentWeapon _unit)}
-    || {!canMove _unit}
+    GVAR(disableAIImediateAction)
     || {_unit getVariable ["ace_captives_isHandcuffed", false]}
     || {_unit getVariable ["ace_captives_issurrendering", false]}
 ) exitWith {false};
```
jokoho48, line 38: this needs to be _stance.

addons/danger/scripts/lambs_danger.fsm
nk3nny: Smart change!

addons/danger/functions/fnc_immediateAction.sqf
```diff
@@ -5,77 +5,87 @@
...
-        "AmovPercMrunSrasWrflDfr_AmovPercMrunSrasWrflDfl"
-    ];
-    _unit switchMove _anim;
-    _unit forceSpeed 4;
-    true
+private _suppression = getSuppression _unit > 0.1;
+private _anim = [];
+
+// move right
+if (_dir > 250 && { random 1 > 0.1 }) then {
```
diwako, line 49: use RND macro

addons/danger/functions/fnc_immediateAction.sqf
```diff
@@ -5,77 +5,87 @@
...
+    if (_suppression) then {
+        _anim append ["FastL", "FastLB"];
+        _unit setUnitPosWeak "DOWN";
+    } else {
+        _anim append ["TactL", "TactLB", "WalkL"];
+    };
+};
+
+// move left
+if (_dir < 80 && { random 1 > 0.1 }) then {
```
diwako, line 60: use RND macro

addons/danger/functions/fnc_indoor.sqf
```diff
@@ -15,4 +15,6 @@
  * Public: No
 */
 params ["_unit"];
-lineIntersects [eyePos _unit, eyePos _unit vectorAdd [0, 0, 4]]
+private _trace = lineIntersectsSurfaces [eyePos _unit, eyePos _unit vectorAdd [0, 0, 4], _unit, objNull, true, 1, "GEOM", "NONE"];
```
diwako, line 18: 4 Meters up will work on most residential buildings, but it will not work on warehouses and the like with higher ceiling

addons/danger/functions/fnc_immediateAction.sqf
jokoho48:
```suggestion
if (_dir > 250 && { RND(0.1) }) then {
```

addons/danger/functions/fnc_immediateAction.sqf
jokoho48:
```suggestion
if (_dir < 80 && { RND(0.1) }) then {
```

addons/danger/functions/fnc_indoor.sqf
jokoho48: i would say 6 meters.

nk3nny: Will fix

jokoho48: i would also change it so that we return all hits and filter them for houses with findIf. so that if you forexample are laying under a car inside a garage it still sees that you are in a building.

jokoho48: so `_trace findIf {_x select 3 isKindOf "Building"};`

PR #124, opened 2020-05-18
Adds enhanced taskAssault

nk3nny:
Adds an improved taskAssault.  
1. Allows multiple taskAssaults to be stacked without too much interference. 
2. taskAssault is more reliable -- functioning more like a proper FORCED MOVEMENT


jokoho48:
taskAssaultUnitReset should properly be renamed to doAssaultUnitReset, naming I followed in the WP module was taskX are user Callable Functions, doX are internal function

PR #126, opened 2020-05-18
Hotfixes for version 2.4.3

nk3nny:
Adds configurable setting for combat mode sharing distance
Fixes vehicles suppressing through terrain
Fixes stringtable description of backpack radios (clearer)


nk3nny:
Reminder to myself: 
Add an exit condition because of max target height over ground for vehicle suppression.

nk3nny:
Reminder to self: 
Add patrol and man static weapons option to both Garrison and Camp

PR #132, opened 2020-06-02
Adds simple reinforcement system

nk3nny:
This branch adds a simple reinforcement framework tied to the share information feature.  It is a feature which is constantly requested on YT and Steam. The branch is built on top of #126 . 

**Function**: 
 Units with the new group variable "lambs_danger_enableGroupReinforce" set to _true_ will be given "MOVE" orders to either the calling units position or the predicted position of the enemy. The destination is determined by having sufficent knowledge. 

The feature merges seamlessly with existing features. Because it only triggers when mission makers _enable_ the setting. It will not disrupt any existing content.  The variable is also enabled for groups set to LAMBS taskPATROL or taskHUNT.  Which makes these scripts much more aggressive and useful.

The groups are given a 'MOVE' order. This will disrupt existing waypoints. However, as these units are now set to be _pseudo_ GUARD units. I believe this is appropriate. 

**Possible problems**
There is no timeout on move orders. So units which receive multiple information instances, possibly with different targets, may become confused and act slower than necessary. Perhaps a simple check for nearby enemies will suffice. 

**Changes**
Adds new group variable "lambs_danger_enableGroupReinforce"
Adds 3DEN group settings for disabling group AI and enabling dynamic reinforcements
Adds setting to taskPatrol and taskHunt

**For which version**
The feature is very simple, but potent and popular. While more invasive than mere hotfixing, I think it can become part of 2.4.3. Lets discuss it however.


addons/danger/functions/fnc_shareInformation.sqf
```diff
@@ -35,25 +35,32 @@ _unit setVariable [QGVAR(currentTask), "Share Information", GVAR(debug_functions
...
+            [_x, [getpos _unit, (_unit targetKnowledge _target) select 6] select (_knowsAbout > 0.5)] remoteExec ["move", leader _x];
+        };
     };
 
-    if ((leader _x) distance _unit < (200 min _range)) then {
-        _x setBehaviour "COMBAT";
-        _x setFormDir ((leader _x) getDir _unit);
+    // set combat mode
+    if ((leader _x) distance2D _unit < ((GVAR(combatShareRange)) min _range) && {!((leader _x) getVariable [QGVAR(disableAI), false])}) then {
+        [_x, "COMBAT"] remoteExec ["setBehaviour", leader _x];
```
jokoho48, line 60: didn't we had an issue with remoteExec on Units and used Groups?

addons/danger/functions/fnc_vehicleAssault.sqf
```diff
@@ -65,8 +66,11 @@ _pos = (eyePos _unit) vectorAdd ((eyePos _unit vectorFromTo (AGLToASL (selectRan
 // look at position
 _vehicle doWatch ASLtoAGL _pos;
 
+// recheck
+if (_vehicle distance (ASLToAGL _pos) < GVAR(minSuppression_range)) exitWith {false};
```
jokoho48, line 70: in L34 you use distance2D and here distance, is that wanted?

jokoho48:
I believe that this feature is to "big" for 2.4.3. we rather should think about moving the goalpost higher and move on to 2.5.0 if you want to have this in the next release.

addons/danger/functions/fnc_shareInformation.sqf
nk3nny: should be fine when run on "leader _x"

addons/danger/functions/fnc_vehicleAssault.sqf
nk3nny: Fixed

PR #138, opened 2020-06-16
Fixed patrol area size in 'garrizon' waypoint (small typo)

DarkWanderer:
This change fixes accidental assignment of `_b` size of patrol area to `_a`


jokoho48:
LGTM

PR #152, opened 2020-08-04
Support for group ownership transfer for "taskX" functions

Lorenz94B:
Hello!
I really like your work and what I would like to be able to do with your mod is transferring groups ownership between headless clients, so that groups that call for taskX functions don't need to remain on the same client. I think that this comes handy in cases where there are a lot of AIs around in dynamic missions with reinforcement spawning, lots of kills etc.

To accomplish this, a simple use of CBA's custom events _should_ do the trick, so I suggest their implementation inside your taskX functions, and from my understandings this should be enough, please correct me if wrong! If I missed something or I did something wrong I really would like to fix it by having a brainstorm together.

Thank you for your time!

When merged this pull request will:
Allow changes in groups ownership between clients for taskX functions, so groups can be transferred on headless clients, for example, freeing the server from (possibly) a lot of AIs.

### DIFF
--- a/addons/wp/XEH_postInit.sqf
+++ b/addons/wp/XEH_postInit.sqf
@@ -1,3 +1,4 @@
 #include "script_component.hpp"
 GVAR(TargetIndex) = 0;
 GVAR(ModuleTargets) = [];
+[QGVAR(taskPatrol), {_this call FUNC(taskPatrol)}] call CBA_fnc_addEventHandler;
--- a/addons/wp/XEH_postInit.sqf
+++ b/addons/wp/XEH_postInit.sqf
@@ -1,16 +1,12 @@
 #include "script_component.hpp"
 GVAR(TargetIndex) = 0;
 GVAR(ModuleTargets) = [];
-
+[QGVAR(taskAssault), {_this call FUNC(taskAssault)}] call CBA_fnc_addEventHandler;
+[QGVAR(taskCQB), {_this call FUNC(taskCQB)}] call CBA_fnc_addEventHandler;
+[QGVAR(taskCamp), {_this call FUNC(taskCamp)}] call CBA_fnc_addEventHandler;
+[QGVAR(taskCreep), {_this call FUNC(taskCreep)}] call CBA_fnc_addEventHandler;
 [QGVAR(taskGarrison), {_this call FUNC(taskGarrison)}] call CBA_fnc_addEventHandler;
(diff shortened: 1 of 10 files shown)


jokoho48:
This PR does not accomplish what you describe in your description. The issue with, not allowing to move the AI from Clients after the function is called, still persist.
If you use Modules(either Zeus or 3den) you already have the same behaviour, such as that the function is already running in a loop as it assumes the AI is still local. Exiting the loop and calling it again on the local machine will not fully work as all the steps beforehand will be lost.

Lorenz94B:
Thank you for your reply! Can you please further explain what you mean with "all the steps beforehand will be lost"? Do you mean evaluations made earlier in the loop (in a previous iteration), or do you mean that without editing also the modules functions the addition I made in the code are useless?

I can't understand if using CBA's custom events can be a solution, but I only integrated it partially (leaving out the modules side of your functions), or not. Maybe if you point me I can make it work, it would be a handy feature in my opinion!

Thank you :)

diwako:
Several task functions have multiple stages to them. One is setting up AI for patrols or putting people in static turrets or teleporting them around. Others like the search tasks run in loops.

> Can you please further explain what you mean with "all the steps beforehand will be lost"

The reason behind "information" getting lost is, none of these stages should run if the unit/group is not local, if you in the midway of the function just exit out and run it on another machine all the same stages will be executed again. This means either what was set up before will be lost or will lead to continuity issues.
CBA events can be used as a glorified remoteExec to run code on other machines, ACE does it, we also do it sometimes. It is not a wrong thing to do. Combined with what was said about the stages and remoteExec, the timing of these need to be correct and these must fully exit out of the function currently running.

As examples:
Most of your changes are done before anything really happens which is correct, however we are already handling that part in our module code, thus it is redundant.
Looking at the task garrison change, you do a targeted event in the end of the patrol stage. This will run the patrol task on another machine; however, it did not exit the script on the machine the event was raised from. This means the script will continue and most likely lead to continuity issues, such as left-over event handlers and seemingly random waypoints.

Lorenz94B:
Thank you for your explanation! I will deeper investigate and try to provide a complete solution, I need to better study the structure of the mod to try again.

diwako:
@Lorenz94B Is this still being worked on?

Lorenz94B:
Hello, I would like to, at the moment I have not much time for Arma (and so will be for the next three weeks).

diwako:
Would you mind us closing this PR in the mean time?

Lorenz94B:
Of course, whatever you think is better! Thank you and have a nice day :)

PR #164, opened 2020-10-01
AIProfiles

jokoho48:


addons/main/CfgAIProfile.hpp
```diff
@@ -0,0 +1,12 @@
+class LAMBS_CfgAIProfiles {
+    class Default {
+        #ifdef ISDEV
+        allowDemoNumberAsBool = 1;
```
diwako, line 4: All of the profile stuff is in the main component, wouldn't it be better to have the dev values in danger and the actual prod default values in main?

addons/main/functions/Profiles/fnc_doesProfileAllow.sqf
```diff
@@ -0,0 +1,41 @@
...
+ * Tactic is allowed to Execute
+ *
+ * Example:
+ * [bob] call lambs_danger_fnc_doesProfileAllow;
+ *
+ * Public: No
+*/
+params [["_target", objNull, [objNull, grpNull]], ["_tactic", "unkown", [""]]];
+
+private _profile = _target getVariable [QGVAR(AIProfile), "default"];
```
diwako, line 20: Kind of a QoL thing. Have a cba setting which profile should be used as default

the CBA setting is a list of available profiles

addons/main/functions/Profiles/fnc_doesProfileAllow.sqf
```diff
@@ -0,0 +1,41 @@
...
+
+private _profileData = GVAR(ProfilesNamespace) getVariable toLower(_profile);
+if (isNil "_profileData") then {
+    _profileData = GVAR(ProfilesNamespace) getVariable "Default";
+};
+_tactic = toLower(_tactic);
+private _value = [_profileData, _tactic] call CBA_fnc_hashGet;
+if (_value isEqualType {}) then {
```
diwako, line 28: I feel like a nil check is needed here for `_value` in the case of some tactic is being requested that is not configured.

addons/main/functions/Profiles/fnc_doesProfileAllow.sqf
jokoho48: a dropdown would be impossible because they can be saved in the description and created via scripted modules. and parsing that while in the editor is from my knowledge not possible. the only thing for CBA setting we could do is a normal text box.

addons/main/functions/Profiles/fnc_doesProfileAllow.sqf
jokoho48: that is already handled by CBA. in the creation of the hash you put in a default return type when it is not defined current Default return is true. ref: https://github.com/nk3nny/LambsDanger/pull/164/files#diff-2f2cb255b4a09218514aa7e856030d9cR23

addons/main/CfgAIProfile.hpp
jokoho48: I would split it into the modules where it is used

addons/main/functions/Profiles/fnc_doesProfileAllow.sqf
diwako: The way I see it, is, a drop down is doable, but only account for config and missionconfig profiles. If someone adds them in editor OR via script it is in their authority to set the default correctly.

addons/main/CfgAIProfile.hpp
diwako: We do need then at least in the wiki some entry with all config entries. It is confusing to look at the code and only see a fraction of possible entries. (Or we have to tell people to use the config viewer)

jokoho48: originally I planned to build a tool that shows and allows you to create and modify profiles in the editor in the release after the profiles where introduced. but writing a small export for the wiki is also fairly easy

PR #173, opened 2021-02-05
WP - Fix task patrol waypoints being at [0,0,0] for local units

PabstMirror:
On local units, waypoints were always placed at [0,0,0]

from biki
>The "Call" in remoteExecCall does not mean that the execution of the remote executed function / command will happen right away

so the logic would be deleted (line 75) before the `getPos` ran inside taskPatrol function


jokoho48:
good catch 👍

PR #186, opened 2021-03-18
Fixes Garrisoned and Camped reinforcements

nk3nny:
- Fixes garrisoned and camped units not being able to reinforce
- Fixes reinforcement variable in taskReset
- Updates stringtable to reflect feature

i.e., before this fix. Units set to waypoint Garrison or Camp were not able to reinforce when the 'Reinforcement' variable was enabled.

### DIFF
--- a/addons/danger/functions/fnc_tacticsReinforce.sqf
+++ b/addons/danger/functions/fnc_tacticsReinforce.sqf
@@ -22,8 +22,13 @@ params ["_unit", ["_target", []], ["_units", []], ["_delay", 300]];
 // exit on dead leader
 if (!(_unit call EFUNC(main,isAlive))) exitWith {false};
 
-// set new time
+// free garrisons
 private _group = group _unit;
+if (EGVAR(main,Loaded_WP) && {!(_unit checkAIFeature "PATH")}) then {
+    _group = [_group] call EFUNC(wp,taskReset);
+};
+
+// set new time
 _group setVariable [QGVAR(enableGroupReinforceTime), time + _delay, true];
 
 // set tasks
--- a/addons/danger/functions/fnc_tacticsReinforce.sqf
+++ b/addons/danger/functions/fnc_tacticsReinforce.sqf
@@ -29,6 +29,7 @@ if (EGVAR(main,Loaded_WP) && {!(_unit checkAIFeature "PATH")}) then {
 };
 
 // set new time
+_group setVariable [QGVAR(enableGroupReinforce), true, true];
 _group setVariable [QGVAR(enableGroupReinforceTime), time + _delay, true];
 
 // set tasks
--- a/addons/danger/functions/fnc_tacticsReinforce.sqf
+++ b/addons/danger/functions/fnc_tacticsReinforce.sqf
@@ -25,7 +25,7 @@ if (!(_unit call EFUNC(main,isAlive))) exitWith {false};
 // free garrisons
 private _group = group _unit;
 if (EGVAR(main,Loaded_WP) && {!(_unit checkAIFeature "PATH")}) then {
-    _group = [_group] call EFUNC(wp,taskReset);
+    _group = [_group, true] call EFUNC(wp,taskReset);
 };
 
 // set new time
--- a/addons/wp/functions/fnc_taskReset.sqf
+++ b/addons/wp/functions/fnc_taskReset.sqf
@@ -33,6 +33,8 @@ private _units = units _group select {!isPlayer _x};
 [_group] call CBA_fnc_clearWaypoints;
 
 // remove LAMBS group variables
+_group setVariable [QEGVAR(danger,disableGroupAI), nil];
+_group setVariable [QEGVAR(danger,enableGroupReinforce), nil, true];
 _group setVariable [QGVAR(taskAssaultDestination), nil, true];
 _group setVariable [QGVAR(taskAssaultMembers), nil, true];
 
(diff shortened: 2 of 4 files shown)


addons/danger/functions/fnc_tacticsReinforce.sqf
```diff
@@ -22,8 +22,13 @@ params ["_unit", ["_target", []], ["_units", []], ["_delay", 300]];
 // exit on dead leader
 if (!(_unit call EFUNC(main,isAlive))) exitWith {false};
 
-// set new time
+// free garrisons
 private _group = group _unit;
+if (EGVAR(main,Loaded_WP) && {!(_unit checkAIFeature "PATH")}) then {
+    _group = [_group] call EFUNC(wp,taskReset);
```
jokoho48, line 28: this may break some missions where a Mission Builders stores the Group for later use.
Task Reset should never be called by us but always by the Mission Builder.

jokoho48 (changes requested):

addons/danger/functions/fnc_tacticsReinforce.sqf
nk3nny: Hmm. Interesting notion.  Of course, the Mission Builder would also be the one to enable "group reinforcements" in the first place.  So that isn't something that just happens. 

I think the feature is important, but it does not have to rely on the nuclear hammer that is taskReset: 
1. A lesser module can be added to Reinforce which specifically frees units from Garrison and Camp. 
2. A modifier to taskReset which _does not_ create a new group.

addons/wp/functions/fnc_taskReset.sqf
```diff
@@ -26,14 +29,8 @@ if (_group isEqualType objNull) then { _group = group _group; };
 // units
 private _units = units _group select {!isPlayer _x};
 
-// make new group + adopt name
-private _groupNew = createGroup [(side _group), true];
-_groupNew setGroupIdGlobal [groupId _group];
-_groupNew setFormation (formation _group);
-
 // remove all current waypoints
 [_group] call CBA_fnc_clearWaypoints;
```
jokoho48, line 33: we should not remove Waypoints on Soft Reset

addons/wp/functions/fnc_taskReset.sqf
```diff
@@ -72,21 +70,37 @@ _group setVariable [QGVAR(taskAssaultMembers), nil, true];
...
-    // rejoin group
-    [_x] joinSilent _groupNew;
-    _x doFollow (leader _x);
+    // rejoin
+    _x doFollow _leader;
 } count _units;
 
+// exit on soft reset
+if (_softReset) exitWith {
+    _units joinSilent _group;
```
jokoho48, line 76: why do we Rejoin the same old group? this i don't think is required here

addons/wp/functions/fnc_taskReset.sqf
```diff
@@ -72,21 +70,37 @@ _group setVariable [QGVAR(taskAssaultMembers), nil, true];
...
+_groupNew setGroupIdGlobal [groupId _group];
+_groupNew setFormation (formation _group);
+
+// rejoin group
+_units joinSilent _groupNew;
+
 // reset lambs variable
 _groupNew setVariable [QEGVAR(danger,disableGroupAI), nil];
-_groupNew setVariable [QEGVAR(danger,enableGroupReinforce), _reinforce, true];
+_groupNew setVariable [QEGVAR(danger,enableGroupReinforce), _group getVariable [QEGVAR(danger,enableGroupReinforce), false], true];
```
jokoho48: shouldn't we also reset those on a soft reset?

jokoho48 (changes requested):

addons/wp/functions/fnc_taskReset.sqf
jokoho48: also what i just noticed enableGroupReinforce should be a main variable and not a danger one if we use it in WP

addons/wp/functions/fnc_taskReset.sqf
nk3nny: I disagree. It isn't much of a reset if the unit continues along its previous orders!

addons/wp/functions/fnc_taskReset.sqf
nk3nny: Its a pain in the ass. but sometimes getting the AI to actually follow the leader seems to require them leaving and rejoining the group. Without it, I found errors in attempting to reset taskCamp.

addons/wp/functions/fnc_taskReset.sqf
nk3nny: 1. I will fix.
2. Lets talk about reinforcement logic before Sunday sometime :)

addons/wp/functions/fnc_taskReset.sqf
jokoho48: the issue with this function is that it destroys placed waypoints in the mission. and this function is getting called automatically. we should never do such a thing. i believe it is against your motto of not interfering with waypoints in the main part of lambs_danger!

nk3nny: You are missing the point.  The "allow reinforcement" variable isn't automatically enabled. The mission maker _MUST_ enable it himself.  Reinforcements isn't something that just happens. No previously made missions will be meaningfully altered.  Furthermore the "taskReset" only triggers for units which have their PATH stopped.  From example our own waypoints or CBA_fnc_TaskDefend. 

Hence. A mission makers that enables BOTH our modules AND the "allow reinforcement" variable can't be said to not willfully be bringing this on himself.

jokoho48: then we should add a warning somewhere that informs the users that using reinforcements may delete Waypoints.

addons/wp/functions/fnc_taskReset.sqf
jokoho48: ok, I see. let's hope this does not break anything

addons/wp/functions/fnc_taskReset.sqf
nk3nny: You might be right. Of course, when we do a hard reset, we make a new group so the concern regarding Waypoints also disappears. 

Suggested compromise:  We add another argument that defaults to false. If set to true, it removes waypoints. @jokoho48

PR #174, opened 2021-02-09
First pass of moving functions

nk3nny:
- Renamed shareInformation and ShareInformationRange
- Moved all doFunctions to Main
- Updated fsms
- Updated variable names

### WHAT THIS DOES

*updated variables*
lambs_main_currentTask
lambs_main_currentTarget
lambs_main_currentTactic -- renamed from lambs_danger_tacticsTask

*Question: I've moved groupMemory*
lambs_main_groupMemory
Arguably this belongs in the danger module, but it is referenced and used in many of the "do" commands just moved into main. I guess in theory it could be added to some waypoints in the future as well?

*Question: What about forceMove?*
Should it be a "main" variable?
lambs_danger_forceMove is used in: 
 L39 findReadyUnits.sqf
 L39 doGroupAssault.sqf

*Question : what about 'dangerRadio' the variable that forces long range radio on them? --> moved to main?* 
NB: This is _set_ in lambs_danger AND SHOULD CONTINUE TO BE SO (I think)
 L40 getShareInformationParam.sqf

*Updated stringtables*

*BUG: getting an error message from settings.sqf*
Getting a weird error on L213

PR #188, opened 2021-03-19
Performance and optimalization

nk3nny:
### This branch does the following:
1. Ruthlessly uses exitWith's in the core brain.sqf to save some cycles.
2. Simplifies tacticsAssess.sqf to streamline assessment phase
3. Simplifies and re-orders some function checks for speed.
4. Moves 'SCREAM' event to 'brainHide.sqf' where investigation and hiding actions should go. (improves performance of brainAssess.sqf which is checked more often!)

_Improvements_
Improves brain.sqf performance by using exitWith checks.
Improves brainEngage.sqf performance by tweaking timers and engagement ranges
Improves brainReact.sqf by simplifying timeout
Improves tacticsAssess.sqf by re-ordering checks and using more exitWiths
Improves tacticsContact.sqf timings and removes unneeded stance reset.
Improves doAssaultCQB.sqf exit conditions
Improves doFleeing.sqf exit conditions
Improves doHide.sqf exit conditions

_Fixes_
Fixes gesture/share information moved to correct state brain.sqf due to fix from #174
Fixes broken smoke throw under tactics_assault in tactics assessment.
Fixes broken wrong variable check in doFleeing.sqf

_Changes_
Changed 'scream'-event now handled by 'brainHide.sqf' instead of 'brainAssess.sqf'

### DIFF
--- a/addons/danger/functions/fnc_tacticsAssess.sqf
+++ b/addons/danger/functions/fnc_tacticsAssess.sqf
@@ -20,12 +20,14 @@
 #define TACTICS_ASSAULT 3
 #define TACTICS_SUPPRESS 4
 #define TACTICS_ATTACK 5
+#define RANGE_MID 220
+#define RANGE_LONG 300
+#define RANGE_THREAT 450
 
 params [["_unit", objNull, [objNull]]];
 
 // check if group AI disabled
 private _group = group _unit;
-if (_group getVariable [QGVAR(disableGroupAI), false]) exitWith {false};
 
 // set variable
 _group setVariable [QGVAR(isExecutingTactic), true];
@@ -36,8 +38,8 @@ _unit setVariable [QEGVAR(main,currentTarget), objNull, EGVAR(main,debug_functio
 _unit setVariable [QEGVAR(main,currentTask), "Tactics Assess", EGVAR(main,debug_functions)];
 
 // gather data
-private _unitCount = count units _unit;     // how many soldiers the leader believes he is leading - nk
-private _enemies = (_unit targets [true, 1200]) select {_unit knowsAbout _x > 1};
+private _unitCount = count units _unit;
+private _enemies = (_unit targets [true, 800]) select {_unit knowsAbout _x > 1};
 private _plan = [];
 
 // leader assess EH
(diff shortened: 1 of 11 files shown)


addons/main/functions/UnitAction/fnc_doHide.sqf
```diff
@@ -21,10 +21,9 @@ params ["_unit", "_pos", ["_range", 55], ["_buildings", []]];
 
 // stopped -- exit
 if (
-    stopped _unit
```
nk3nny, line 24: Stopped is a legacy feature from older versions. https://community.bistudio.com/wiki/stop . I thought it caught "doStop" but it didn't. Only place I've seen it use is in the Civilian FSM script.

addons/danger/functions/fnc_tacticsContact.sqf
```diff
@@ -46,15 +46,14 @@ _group setVariable [QEGVAR(main,currentTactic), "Contact!", EGVAR(main,debug_fun
...
-            {_x setUnitPosWeak "AUTO"} foreach units _group;
             private _leader = leader _group;
             if (_leader call FUNC(isLeader)) then {
                 [_leader, _leader findNearestEnemy _leader] call FUNC(tactics);
             };
         };
     },
     [_group, attackEnabled _group],
-    _delay + random 8
+    _delay + random 12
```
nk3nny, line 56: Increased the random spread a bit to spread out leadership checks

addons/danger/functions/fnc_tacticsAssess.sqf
```diff
@@ -20,12 +20,14 @@
 #define TACTICS_ASSAULT 3
 #define TACTICS_SUPPRESS 4
 #define TACTICS_ATTACK 5
+#define RANGE_MID 220
+#define RANGE_LONG 300
+#define RANGE_THREAT 450
```
nk3nny, line 25: These should have been in before.  Through they could be transitioned to settings. I propose simply Defining for now. 
RANGE_THREAT points to both armored vehicles and aircraft.

jokoho48: if we use Defines in Multiple Areas we should move them to the script_component.hpp in the highest common level

nk3nny: Good point. Tempting to do so now, but I feel this pull request is already dangerously sprawling.

nk3nny: In fact: What about adding CBA sliders? Allowing customization.  Much akin to the CQB variable.  Something for 2.5+ methinks.

jokoho48 (approved):

PR #190, opened 2021-03-24
Brain performance and timings

nk3nny:
### Brain performance and tuned timings: 
Adds exit condition on 'lambs_danger_forceMove' in brainForced
Improved performance of brainEngage by having zero-time exit variables
Improved performance by skipping unnecessary vector additions to suppression

### DIFF
--- a/addons/danger/functions/fnc_isForced.sqf
+++ b/addons/danger/functions/fnc_isForced.sqf
@@ -10,17 +10,12 @@
  * bool
  *
  * Example:
- * [bob] call lambs_danger_fnc_isForced;
+ * bob call lambs_danger_fnc_isForced;
  *
  * Public: No
 */
-params ["_unit"];
-fleeing _unit
-|| {_unit getVariable [QGVAR(disableAI), false]}
-|| {_unit getVariable [QGVAR(forceMove), false]}
-|| {currentCommand _unit in ["ATTACK", "GET IN", "ACTION", "HEAL", "REARM", "JOIN"]}
-|| {!(_unit call EFUNC(main,isAlive))}
-|| {(_unit getVariable ["ace_medical_ai_healQueue", []]) isNotEqualTo []}
-|| {GVAR(disableAIPlayerGroup) && {isPlayer leader _unit}}
-|| {(behaviour _unit) isEqualTo "CARELESS"}
-|| {!(_unit checkAIFeature "MOVE")}
+_this getVariable [QGVAR(forceMove), false]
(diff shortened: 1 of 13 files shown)


addons/main/script_macros.hpp
```diff
@@ -18,7 +18,7 @@
     _unit \
 }
 
-// #define DISABLE_COMPILE_CACHE
+#define DISABLE_COMPILE_CACHE
```
jokoho48, line 21: you forgot this here D:

nk3nny: Haha, yeah noticed. blargh

addons/danger/scripts/lambs_danger.fsm
```diff
@@ -230,7 +245,7 @@ class FSM
                         itemno = 9;
                         init = /*%FSM<STATEINIT""">*/"" \n
                          "// keep attacking" \n
-                         "if (_dangerCause in [8, 9] && {_dangerCausedBy call lambs_main_fnc_isAlive}) then {" \n
+                         "if (_dangerCause in [8, 9] && {alive _dangerCausedBy}) then {" \n
```
diwako, line 248: Why the change to just a simple alive check?

nk3nny: Two reasons: 
1. Performance (slight)
2. ACE3

ACE3 sets incapacitated units as functionally _invisible_ to the AI. However, that invisibility does not mean the AI isn't being told by its native FSM to try to find and shoot them.  Hence, a simple checks allows our dangerFSM to run _on_ that target-- instead of having them uselessly flail about. 

As it pertains to vehicles, this means a range of interesting and immersive actions.  Such as generally suppressing the location of suspected enemies and _jinking_ out of danger of the same. :)

addons/danger/functions/fnc_isForced.sqf
```diff
@@ -10,17 +10,12 @@
...
  *
  * Example:
- * [bob] call lambs_danger_fnc_isForced;
+ * bob call lambs_danger_fnc_isForced;
  *
  * Public: No
 */
-params ["_unit"];
-fleeing _unit
-|| {_unit getVariable [QGVAR(disableAI), false]}
```
jokoho48, line 19: do we deprecating this variable? or is it handled elsewhere?

jokoho48 (approved):

addons/danger/scripts/lambs_danger.fsm
nk3nny: Reverted

PR #193, opened 2021-03-24
Remove the Draw EventHandler when it is not required

jokoho48:

1. *Title*

### DIFF
--- a/addons/main/settings.sqf
+++ b/addons/main/settings.sqf
@@ -221,6 +221,16 @@ _curCat = LSTRING(Settings_Debug);
 
         GVAR(debug_drawRectCacheCurator) = [];
         GVAR(debug_drawRectInUseCurator) = [];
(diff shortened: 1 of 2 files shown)


addons/main/settings.sqf
```diff
@@ -221,6 +221,16 @@ _curCat = LSTRING(Settings_Debug);
 
         GVAR(debug_drawRectCacheCurator) = [];
         GVAR(debug_drawRectInUseCurator) = [];
+        if (GVAR(debug_Drawing)) then {
```
diwako, line 224: Wouldn't it be cleaner to use the execution blocks `params ["_value"];` here?

jokoho48: I forgot how the parameters of the settings callback worked

jokoho48:
done in #195

PR #194, opened 2021-03-27
Remove doFire from tacticsContacts

diwako:

1. Make AI stop shoot friendlies because they are annoyed by them firing at nothing

### DIFF
--- a/addons/danger/functions/fnc_tacticsContact.sqf
+++ b/addons/danger/functions/fnc_tacticsContact.sqf
@@ -109,11 +109,10 @@ if (_deadOrSuppressed isEqualTo -1 && {_unit distance2D _enemy < (GVAR(cqbRange)
         private _distanceAssault = RND(0.2) && {_x distance2D _enemy < GVAR(cqbRange)};
         if (_distanceAssault) then {
             [_x, _enemy] call EFUNC(main,doAssault);
-            _x doFire (vehicle _enemy);
         } else {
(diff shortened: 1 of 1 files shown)


addons/danger/functions/fnc_tacticsContact.sqf
```diff
@@ -109,11 +109,10 @@ if (_deadOrSuppressed isEqualTo -1 && {_unit distance2D _enemy < (GVAR(cqbRange)
         private _distanceAssault = RND(0.2) && {_x distance2D _enemy < GVAR(cqbRange)};
         if (_distanceAssault) then {
             [_x, _enemy] call EFUNC(main,doAssault);
-            _x doFire (vehicle _enemy);
         } else {
             [_x, ATLtoASL ((_unit getHideFrom _enemy) vectorAdd [0, 0, 0.3 + random 1])] call EFUNC(main,doSuppress);
         };
-    } foreach _units;
+    } foreach (_units select {currentCommand _x isEqualTo ""});
```
diwako, line 115: Please discuss this line

jokoho48: I don't know what the results are of that filtering. I think Kenny should decide if that is important

nk3nny: Interesting tweak. It means that the only units affected will be the ones not already moving.  Probably a good tweak in this case.

nk3nny (approved):

PR #201, opened 2021-03-28
Add custom share information handlers

mjc4wilton:

1. Add `lambs_main_fnc_addShareInformationHandler` which allows for custom functions to block information sharing (similar to how `ace_explosives_fnc_addDetonateHandler` [works](https://ace3mod.com/wiki/framework/explosives-framework.html#53-detonation-handler)).
2. Add myself to contributor list.

### DIFF
--- a/addons/main/functions/fnc_addShareInformationHandler.sqf
+++ b/addons/main/functions/fnc_addShareInformationHandler.sqf
@@ -1,8 +1,8 @@
 /*
  * Author: mjc4wilton
  * Adds a custom handler to be called whenever the AI attempt to share information.
- * Code should return a BOOL where true allows information to be passed and false blocks it.
- * A single false will cause information to not be shared.
+ * Code should return a BOOL where false allows information to be passed and true blocks it.
+ * A single true will cause information to not be shared.
  *
  * Arguments:
  * 0: function <CODE>
--- a/addons/main/functions/fnc_addShareInformationHandler.sqf
+++ b/addons/main/functions/fnc_addShareInformationHandler.sqf
@@ -1,0 +1,27 @@
+/*
+ * Author: mjc4wilton
+ * Adds a custom handler to be called whenever the AI attempt to share information.
+ * Code should return a BOOL where true allows information to be passed and false blocks it.
+ * A single false will cause information to not be shared.
+ *
+ * Arguments:
+ * 0: function <CODE>
+ *
+ * Arguments passed to function:
+ * 0: unit sharing information <OBJECT>
+ * 1: enemy target <OBJECT>
+ * 2: range to share information, default 350 <NUMBER>
+ * 3: override radio ranges, default false <BOOLEAN>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [{false}] call lambs_main_fnc_addShareInformationHandler;
+ *
+ * Public: No
+*/
+
+params [["_code", {true}, [{}]]];
+
+GVAR(shareHandlers) pushBack _code;
--- a/addons/main/functions/fnc_addShareInformationHandler.sqf
+++ b/addons/main/functions/fnc_addShareInformationHandler.sqf
@@ -22,6 +22,6 @@
  * Public: No
 */
 
-params [["_code", {true}, [{}]]];
+params [["_code", {false}, [{}]]];
 
 GVAR(shareHandlers) pushBack _code;
--- a/addons/main/functions/fnc_doShareInformation.sqf
+++ b/addons/main/functions/fnc_doShareInformation.sqf
@@ -32,6 +32,14 @@ if (isNull _target) then {
     _target = _unit findNearestEnemy _unit;
 };
 
+// custom handlers
+private _handlersReturn = true;
+{
+    private _handlerResult = [_unit, _target, _range, _override] call _x;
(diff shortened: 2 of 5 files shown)


addons/main/functions/fnc_addShareInformationHandler.sqf
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Example:
+ * [{false}] call lambs_main_fnc_addShareInformationHandler;
+ *
+ * Public: No
+*/
+
+params [["_code", {true}, [{}]]];
+
+GVAR(shareHandlers) pushBack _code;
```
diwako, line 27: Where is `GVAR(shareHandlers)` being initialized?

mjc4wilton: Knew I was forgetting something. Just pushed up a define in preInit.

jokoho48:
I believe this is basically what we are already playing todo in https://github.com/nk3nny/LambsDanger/pull/164

mjc4wilton:
> I believe this is basically what we are already playing todo in #164

I didn't look too hard into the addon / current PR's when writing this. This was basically just because I needed some way of doing this for an Electronic Warfare mod I'm writing to enable ACRE / TFAR jammers to work on AI.

addons/main/functions/fnc_doShareInformation.sqf
```diff
@@ -32,6 +32,14 @@ if (isNull _target) then {
     _target = _unit findNearestEnemy _unit;
 };
 
+// custom handlers
+private _handlersReturn = true;
+{
+    private _handlerResult = [_unit, _target, _range, _override] call _x;
+    if (_handlerResult isEqualTo false) exitWith {_handlersReturn = false};
```
jokoho48, line 39: as far as I know, this function may be called in a Scheduled Environment what would if the _x function returns nil throw a script error here.

jokoho48 (changes requested):

addons/main/functions/fnc_doShareInformation.sqf
```diff
@@ -32,6 +32,14 @@ if (isNull _target) then {
...
+} forEach GVAR(shareHandlers);
+if (!_handlersReturn) exitWith {false};
```
jokoho48, line 41: same as above.

addons/main/functions/fnc_doShareInformation.sqf
diwako: Is it also necessary to write the result into a variable?

mjc4wilton: > as far as I know, this function may be called in a Scheduled Environment what would if the _x function returns nil throw a script error here.

This is why isEqualTo is being utilized here. Nil returns or anything other than false would basically be treated as a true.

> Is it also necessary to write the result into a variable?

Strictly speaking, no, but it makes it a bit more readable.

jokoho48: isEqualTo is no fix for that because nil is handled as undefined and will throw an `Error Undefined variable in expression`.

mjc4wilton: Should be handled with recent commit

addons/main/functions/fnc_doShareInformation.sqf
```diff
@@ -32,6 +32,14 @@ if (isNull _target) then {
     _target = _unit findNearestEnemy _unit;
 };
 
+// custom handlers
+private _handlersReturn = true;
+{
+    private _handlerResult = [_unit, _target, _range, _override] call _x;
+    if (!isNil _handlerResult && _handlerResult isEqualTo false) exitWith {_handlersReturn = false};
+} forEach GVAR(shareHandlers);
+if (!_handlersReturn) exitWith {false};
```
diwako, line 41: May I suggest negating this? The idea can be borrowed from the vanilla event handlers which allow to overwrite the handles status. Meaning returning `true` means it was handled and no further need to do something and `false` means "not handled, continue". Of course the header of the `addShareInformationHandler` functions needs to change a little.

Another change suggestion is to wrap the second condition in a code block so it actually does lazy eval.

```suggestion
private _stopShare = false;
{
    private _callbackResult = [_unit, _target, _range, _override] call _x;
    if (!(isNil "_callbackResult") && {_callbackResult isEqualTo true}) exitWith {_stopShare = true};
} forEach GVAR(shareHandlers);
if (_stopShare) exitWith {false};
```

jokoho48 (approved):

addons/main/functions/fnc_addShareInformationHandler.sqf
```diff
@@ -0,0 +1,27 @@
...
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [{false}] call lambs_main_fnc_addShareInformationHandler;
+ *
+ * Public: No
+*/
+
+params [["_code", {true}, [{}]]];
```
diwako, line 25: With 2.5.0 released this can be merged now. I am requesting just one small change since everything else was already inverted.

```suggestion
params [["_code", {false}, [{}]]];
```

diwako (changes requested):

diwako (approved):

PR #209, opened 2021-04-16
Improves isNight check performance by RCA3

nk3nny:
Improves isNight.sqfo performance in main by use of getEnvSound Controllers.
Improvement suggested by RCA3 on BIS forums. 

https://community.bistudio.com/wiki/getEnvSoundController


diwako (approved):
Just tried it out on a custom map with environment disabled. Seems to work, is also faster than the function call and such.

jokoho48 (approved):

PR #210, opened 2021-04-17
Improve AI Artillery -- disembarking and improving doArtillery

nk3nny:
Adds artillery crew will disembark when threatened by near enemies
Adds skipCheckRounds will also skip main strike delay*
Adds debug marker for spotter
Fixes check round barrages nearly always be one round
Fixes debug markers were occasionally not removed when caller died
Changes artillery debug markers uses a colour not used by a faction
Changes: static weapon crew will disembark when threatened (instead of only effective commander)
Minor header and variable name clean ups

*This is done according to earlier discussion and the actual description given in stringtables!

### DIFF
--- a/addons/danger/functions/fnc_brainVehicle.sqf
+++ b/addons/danger/functions/fnc_brainVehicle.sqf
@@ -57,6 +57,12 @@ private _attack = _cause in [DANGER_ENEMYDETECTED, DANGER_ENEMYNEAR, DANGER_HIT,
 private _artillery = _vehicle getVariable [QEGVAR(main,isArtillery), getNumber (configOf _vehicle >> "artilleryScanner") > 0];
 if (_artillery) exitWith {
     _vehicle setVariable [QEGVAR(main,isArtillery), true];
+
+    // enemies within 12-30m may cause crew to disembark!
+    if (_attack && {_dangerCausedBy distanceSqr _vehicle < (144 + random 324)} && {currentCommand _unit isEqualTo ""}) then {
+        (units _unit) orderGetIn false;
+        _unit setSuppression 0.94; // to prevent instant laser aim on exiting vehicle
+    };
     [_timeout] + _causeArray
 };
 
@@ -69,7 +75,7 @@ if (_static) exitWith {
 
     // get out if enemy near
     if ((_unit findNearestEnemy _dangerPos) distance _vehicle < (6 + random 15)) then {
-        [_unit] orderGetIn false;
+        (units _unit) orderGetIn false;
         _unit setSuppression 0.94; // to prevent instant laser aim on exiting vehicle
     };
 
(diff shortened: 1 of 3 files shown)


addons/wp/functions/fnc_doArtillery.sqf
```diff
@@ -135,15 +145,15 @@ if (canFire _gun && {_caller call EFUNC(main,isAlive)}) then {
...
-            {
-                {deleteMarker _x; true} count _this
-            }, _mList, 60
-        ] call CBA_fnc_waitAndExecute;
-    };
+// clean markers!
+if (_markerList isNotEqualTo []) then {
+    [
+        {
+            {deleteMarker _x; true} count _this
```
diwako, line 154: Can we make a standard forEach out of this?

addons/danger/functions/fnc_brainVehicle.sqf
```diff
@@ -57,6 +57,12 @@ private _attack = _cause in [DANGER_ENEMYDETECTED, DANGER_ENEMYNEAR, DANGER_HIT,
 private _artillery = _vehicle getVariable [QEGVAR(main,isArtillery), getNumber (configOf _vehicle >> "artilleryScanner") > 0];
 if (_artillery) exitWith {
     _vehicle setVariable [QEGVAR(main,isArtillery), true];
+
+    // enemies within 12-30m may cause crew to disembark!
+    if (_attack && {_dangerCausedBy distanceSqr _vehicle < (144 + random 324)} && {currentCommand _unit isEqualTo ""}) then {
```
diwako, line 62: When using the squared distance check, don't you need to compare it to a squared distance as well. so 144^2?

nk3nny: 144 is 12 * 12 

;)

addons/wp/functions/fnc_doArtillery.sqf
nk3nny: Wilco

addons/wp/functions/fnc_doArtillery.sqf
```diff
@@ -49,47 +50,50 @@ if !((vehicle _gun) isKindOf "StaticMortar") then {
...
+            // randomize target location
+            private _target = _center getPos [(100 + (random _accuracy * 2)) / _check, _direction + 45 - random 90];
+
+            private _ammo = (getArtilleryAmmo [_gun]) param [0, ""];
+            if (_ammo isEqualTo "") exitWith {};
+
+            if (_target inRangeOfArtillery [[_gun], _ammo]) then {
+
+                // check rounds barrage
+                _gun commandArtilleryFire [_target, _ammo, 1 + random 2];
```
jokoho48, line 86: This may create issues with ACE_CSW because commandArtilleryFire only fires maximum the amounts of rounds left in the magazine, and with ACE_CSW that is always limited to 1. I think we need to keep it looping manually as it was

nk3nny: for loop re-implemented.

jokoho48 (approved):

PR #222, opened 2021-04-21
Simplified Params array in fsm

nk3nny:
As per discussion. Simplified Params array in FSM.  Removed unnecessary default entries


jokoho48 (approved):

diwako (approved):
Was worried that one of the values could be nil, then check fnc_brain to confirm it is not the case.

PR #224, opened 2021-04-21
add formations module

jokoho48:

1. *Title*


diwako (approved):

addons/formations/config.cpp
```diff
@@ -0,0 +1,15 @@
+#include "script_component.hpp"
+class CfgPatches {
+    class ADDON {
+        name = COMPONENT_NAME;
+        units[] = {};
+        weapons[] = {};
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = {"lambs_main"};
```
nk3nny, line 8: Do we need to add more required addons to this? "A3_Data_F" ?

jokoho48: lambs_main requires cba_main what requires cba_common what requires A3_Data_F_AoW_Loadorder what requires all vanilla classes afaik

nk3nny (approved):

PR #227, opened 2021-04-21
Remove assault stance changes

nk3nny:
Uncomments stance changes on aggressive (assault) actions.  This reduces the 'enforced' stance changes considerably.  The tweaks works positively by keeping the majority of stance changes on group or defensive actions.  I believe this works with better synergy with our new formations/pathfinding module. 

Fixes units standing up at inopportune times. 

I would like to test this publicly in the SW 2.5.1 'RC'. The features can be easily re-implemented if players experience AI trying to crawl inside buildings in CQB.  In my own testing. That has not tended to occur.

### DIFF
--- a/addons/main/functions/UnitAction/fnc_doAssault.sqf
+++ b/addons/main/functions/UnitAction/fnc_doAssault.sqf
@@ -54,7 +54,7 @@ private _pos = if (_buildings isEqualTo []) then {
 };
 
 // stance and speed
-_unit setUnitPosWeak selectRandom ["UP", "UP", "MIDDLE"];
+//_unit setUnitPosWeak selectRandom ["UP", "UP", "MIDDLE"];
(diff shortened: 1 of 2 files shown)


jokoho48 (approved):

addons/main/functions/UnitAction/fnc_doAssault.sqf
```diff
@@ -54,7 +54,7 @@ private _pos = if (_buildings isEqualTo []) then {
 };
 
 // stance and speed
-_unit setUnitPosWeak selectRandom ["UP", "UP", "MIDDLE"];
+//_unit setUnitPosWeak selectRandom ["UP", "UP", "MIDDLE"];
```
diwako, line 57: Do you intent to re-add those? If not why not delete the whole line?

PR #230, opened 2021-04-23
Improve and optimise debug draw

jokoho48:

1. *Adds more detail to debugDraw Cards*
2. *Cleans up some values that are removed when not set*
3. *Fixes issues that Cards did get drawn at ground level when AI is in building*
4. *the `GROUP LEADER` text is now in the side color*
5. *General Code and Performance improvements*

### DIFF
--- a/addons/main/functions/debug/fnc_debugDraw.sqf
+++ b/addons/main/functions/debug/fnc_debugDraw.sqf
@@ -122,14 +122,14 @@ private _fnc_debug_drawRect = {
 private _sideUnknownColor = GVAR(debug_sideColorLUT) get sideUnknown;
 {
     private _unit = _x;
-    private _renderPos = getPosVisual _unit;
-    if (((positionCameraToWorld [0, 0, 0]) distance _renderPos) <= 1000) then {
+    private _renderPos = getPosATLVisual _unit;
+    if (((positionCameraToWorld [0, 0, 0]) distance _renderPos) <= viewDistance) then {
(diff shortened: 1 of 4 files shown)


nk3nny (approved):

addons/main/functions/debug/fnc_debugDraw.sqf
```diff
@@ -119,42 +119,49 @@ private _fnc_debug_drawRect = {
...
 };
 
+private _sideUnknownColor = GVAR(debug_sideColorLUT) get sideUnknown;
 {
     private _unit = _x;
-    private _renderPos = getPosVisual _unit;
-    if (((positionCameraToWorld [0, 0, 0]) distance _renderPos) <= 1000) then {
-    // if (true) then {
+    private _renderPos = getPosATLVisual _unit;
+    if (((positionCameraToWorld [0, 0, 0]) distance _renderPos) <= viewDistance) then {
```
diwako, line 126: Minor improvement to move the call of `viewDistance` out of the loop into some private variable?

addons/main/functions/debug/fnc_debugDraw.sqf
```diff
@@ -119,42 +119,49 @@ private _fnc_debug_drawRect = {
...
+        if (!GVAR(debug_drawAllUnitsInVehicles) && {_unit isNotEqualTo (effectiveCommander (vehicle _unit))}) exitWith {};
```
diwako, line 127: Why not include that into the if right above this line?

PR #231, opened 2021-04-24
Fix retreat option in task assault

nk3nny:
### Bugfix
Fixes units 'walking' when retreat option is enabled in taskAssault
Adds minor performance improvement to animation selection


jokoho48 (approved):

addons/wp/functions/fnc_doAssaultUnitReset.sqf
```diff
@@ -23,8 +23,8 @@ _unit setVariable [QEGVAR(danger,forceMove), nil];
 // stance
 _unit forceSpeed -1;
 _unit setUnitPos "AUTO";
-_unit doMove getPosASL _unit;
-_unit doFollow leader _unit;
+_unit doMove (getPosWorld _unit);
```
diwako, line 26: Isn't doMove using ATL as input?

diwako (approved):

PR #232, opened 2021-04-24
Fixes excessive use of AI smoke

nk3nny:
### When merged this pull request will: Reduce smoke usage
- Require AI to be suppressed
- Remove duplicate smoke usage in tacticsAssess

Answers feedback from a number of sources. Latest being discord.

### DIFF
--- a/addons/danger/functions/fnc_tacticsAssault.sqf
+++ b/addons/danger/functions/fnc_tacticsAssault.sqf
@@ -77,13 +77,13 @@ _group enableAttack false;
 [_unit, "combat", "Advance", 125] call EFUNC(main,doCallout);
 
 // concealment
-if (_unit distance2D _target > 15) then {
+if (_unit distance2D _target > 25) then {
 
     // leader smoke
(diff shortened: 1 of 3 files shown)


jokoho48 (approved):

addons/danger/functions/fnc_tacticsAssault.sqf
```diff
@@ -77,13 +77,13 @@ _group enableAttack false;
 [_unit, "combat", "Advance", 125] call EFUNC(main,doCallout);
 
 // concealment
-if (_unit distance2D _target > 15) then {
+if (_unit distance2D _target > 25) then {
 
     // leader smoke
-    [_unit, _target] call EFUNC(main,doSmoke);
+    if ((getSuppression _unit) isNotEqualTo 0) then {[_unit, _target] call EFUNC(main,doSmoke);};
```
diwako, line 83: Not sure if that will help. Means they will use smoke even if someone fires one shot at them. Probably a threshold like 0.1 needs to be set here.

PR #233, opened 2021-04-24
Refines Suppression in brainEngage and tacticsAssess

nk3nny:
**Refines individual suppression and engagement in three ways:**
1. Makes it possible to quit engagement cycle by setting combatMode to "BLUE" or "GREEN".  I.e., ordering units to hold fire.
2. Halves chance of suppression when enemy is merely 'detected' at not actually considered a target easy to fire at.
3. Removes needReload which wasn't all that successful and also quite slow.

**Refines tacticsAssess:**
* Adds and applies "RANGE_NEAR" define
* Removes Suppression chance when combatMode is "white" -- i.e., group is supposed to hold fire.
* Minor performance improvements by using getPosWorld and getPosATL where relevant

Changes in both functions make the AI less likely to become fully stuck in a suppression loop. Particularly when the enemy isn't actually spotted.

### DIFF
--- a/addons/danger/functions/fnc_brainEngage.sqf
+++ b/addons/danger/functions/fnc_brainEngage.sqf
@@ -37,13 +37,13 @@ if (
     isNull _target
     || {_unit knowsAbout _target isEqualTo 0}
     || {(weapons _unit) isEqualTo []}
-    || {needReload _unit > 0.8}
+    || {combatMode _unit in ["BLUE", "GREEN"]}
 ) exitWith {
     _timeout
 };
 
-// look at_this
-if (_unit knowsAbout _target > 3.9) then {
+// look at target
+if ((_unit knowsAbout _target) isEqualTo 4) then {
     _unit lookAt _target;
 };
 
@@ -68,7 +68,7 @@ _unit forceSpeed ([-1, 1] select (_type isEqualTo DANGER_CANFIRE));
 if (
     _distance < 500
     && {RND(getSuppression _unit)}
-    && {_type in [DANGER_ENEMYDETECTED, DANGER_CANFIRE]}
+    && {_type isEqualTo DANGER_CANFIRE || {RND(0.5) || {_type isEqualTo DANGER_ENEMYDETECTED}}}
 ) exitWith {
     [_unit, ATLtoASL ((_unit getHideFrom _target) vectorAdd [0, 0, 0.5])] call EFUNC(main,doSuppress);
     _timeout + 4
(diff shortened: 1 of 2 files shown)


addons/danger/functions/fnc_brainEngage.sqf
```diff
@@ -37,13 +37,13 @@ if (
     isNull _target
     || {_unit knowsAbout _target isEqualTo 0}
     || {(weapons _unit) isEqualTo []}
-    || {needReload _unit > 0.8}
+    || {combatMode _unit in ["BLUE", "GREEN"]}
```
diwako, line 40:
```suggestion
    || {(combatMode _unit) in ["BLUE", "GREEN"]}
```
;)

addons/danger/functions/fnc_tacticsAssess.sqf
```diff
@@ -102,7 +103,7 @@ if !(_enemies isEqualTo [] || {_unitCount < random 3}) then {
     // enemies within X meters of leader and either attacker or unit is inside buildings
     private _inside = _unit call EFUNC(main,isIndoor);
     private _nearIndoorTarget = _enemies findIf {
-        _unit distance2D _x < (GVAR(cqbRange) * 1.4)
+        _unit distance2D _x < RANGE_NEAR
```
diwako, line 106: Are we intentionally removing configurability here?

addons/danger/functions/fnc_tacticsAssess.sqf
```diff
@@ -118,7 +119,7 @@ if !(_enemies isEqualTo [] || {_unitCount < random 3}) then {
         !_speedMode
         && {_unit distance2D _x > RANGE_LONG}
         && {_unit knowsAbout _x < 2}
-        && {((getPosASL _x) select 2) > ((_eyePos select 2) + 15)}
+        && {((getPosWorld _x) select 2) > ((_eyePos select 2) + 15)}
```
diwako, line 122: Eyepos reports ASL back. This will become an issue over water.

nk3nny: Perhaps exit all tactics over or in water?

addons/danger/functions/fnc_tacticsAssess.sqf
nk3nny: Yes. 
I feel that layers and layers of hidden interaction with the CQB Range setting is probably less clear than it should be. I haven't checked the profile changes yet, but I suspect that sort of configuration for _group tactics_ is better handled there.

addons/danger/functions/fnc_tacticsAssess.sqf
jokoho48: when a object is place over water that position may also be water is the issue(for example the piers are sometimes overwater)

jokoho48 (approved):

PR #235, opened 2021-04-24
Minor fixes to language in waypoints

nk3nny:
### Language fixes to waypoint description
- Fixes missing taskAssault - Retreat mode waypoint text
- Minor tweaks to other waypoint text.
- Removes reference to 'unit' or 'units' to maintain consistency.


jokoho48:
i feel like we should move that to the Stringtable

PR #242, opened 2021-05-03
Adds player check to units listing in tacticsContact

nk3nny:
Fixes player led groups failing to reform on player after shooting starts.

### DIFF
--- a/addons/danger/functions/fnc_tacticsContact.sqf
+++ b/addons/danger/functions/fnc_tacticsContact.sqf
@@ -62,7 +62,7 @@ _group setFormDir (_unit getDir _enemy);
 
 // gesture + callouts for larger units
 private _stealth = (behaviour _unit) isEqualTo "STEALTH";
-private _units = (units _unit) select {(currentCommand _x) in ["", "MOVE"]};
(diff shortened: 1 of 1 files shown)


addons/danger/functions/fnc_tacticsContact.sqf
```diff
@@ -62,7 +62,7 @@ _group setFormDir (_unit getDir _enemy);
 
 // gesture + callouts for larger units
 private _stealth = (behaviour _unit) isEqualTo "STEALTH";
-private _units = (units _unit) select {(currentCommand _x) in ["", "MOVE"]};
+private _units = (units _unit) select {(currentCommand _x) in ["", "MOVE"] && {!isPlayer _x}};
```
jokoho48, line 65: i would say flip that around checking for isPlayer should be faster than the in here
```suggestion
private _units = (units _unit) select {!(isPlayer _x) && {(currentCommand _x) in ["", "MOVE"]}};
```

nk3nny: First option is marginally faster when testing 30+ units doing various actions.  I guess this might be because isPlayer is fairly slow?  Could you check on your end to see if that can be replicated

PR #249, opened 2021-05-09
initial suppressive fire check

diwako:

1. Adds suppression LOS check function
2. Use LOS check in doSuppression function
3. Enable the check in several tactics

- [x] check and account for missing p3d ending in model config

### DIFF
--- a/addons/danger/functions/fnc_tacticsSuppress.sqf
+++ b/addons/danger/functions/fnc_tacticsSuppress.sqf
@@ -23,7 +23,7 @@ params ["_unit", "_target", ["_units", []], ["_delay", 17]];
 _target = _target call CBA_fnc_getPos;
 
 // exit with flank squad leader cannot suppress from here
-if !([_unit, _target] call FUNC(shouldSuppressPosition)) exitWith {
+if !([_unit, ATLToASL _target] call EFUNC(main,shouldSuppressPosition)) exitWith {
     [_unit, _target] call FUNC(tacticsFlank);
 };
 
--- a/addons/danger/functions/fnc_tacticsSuppress.sqf
+++ b/addons/danger/functions/fnc_tacticsSuppress.sqf
@@ -23,7 +23,7 @@ params ["_unit", "_target", ["_units", []], ["_delay", 17]];
 _target = _target call CBA_fnc_getPos;
 
 // exit with flank squad leader cannot suppress from here
-if !([_unit, ATLToASL _target] call EFUNC(main,shouldSuppressPosition)) exitWith {
+if !([_unit, (ATLToASL _target) vectorAdd [0, 0, 2.5]] call EFUNC(main,shouldSuppressPosition)) exitWith {
     [_unit, _target] call FUNC(tacticsFlank);
 };
 
--- a/addons/danger/functions/fnc_tacticsSuppress.sqf
+++ b/addons/danger/functions/fnc_tacticsSuppress.sqf
@@ -22,12 +22,8 @@ params ["_unit", "_target", ["_units", []], ["_delay", 17]];
 // find target
 _target = _target call CBA_fnc_getPos;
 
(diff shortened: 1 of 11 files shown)


addons/danger/functions/fnc_tacticsSuppress.sqf
```diff
@@ -22,12 +22,8 @@ params ["_unit", "_target", ["_units", []], ["_delay", 17]];
...
 _target = _target call CBA_fnc_getPos;
 
-// exit with flank if visible distance is 71+ meters from actual target
-private _targetASL = ATLtoASL _target;
-private _vis = lineIntersectsSurfaces [(getPosWorld _unit) vectorAdd [0, 0, 3], _targetASL vectorAdd [0, 0, 3], _unit, objNull, true, 3, "FIRE", "GEOM"];
-private _index = _vis findIf {(_x select 2) isKindOf "Building"};
-
-if (_index isNotEqualTo -1 && {((_vis select _index) select 0) vectorDistanceSqr _targetASL > 5041}) exitWith {
+// exit with flank squad leader cannot suppress from here
+if !([_unit, _target] call FUNC(shouldSuppressPosition)) exitWith {
```
nk3nny, line 26: EFUNC(main,shouldSuppressPosition)

blah blah

addons/main/functions/fnc_shouldSuppressPosition.sqf
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Example:
+ * [bob, angryJoe] call lambs_main_fnc_shouldSuppressPosition;
+ *
+ * Public: No
+*/
+params ["_unit", "_target"];
+
+private _targetASL = ATLtoASL (_target call CBA_fnc_getPos);
+private _unitASL = getPosASL _unit;
```
nk3nny, line 21: Consider using eyePos and a smaller added vector to account for prone units?

addons/main/XEH_postInit.sqf
```diff
@@ -10,7 +10,7 @@ GVAR(CalloutCacheNamespace) = call CBA_fnc_createNamespace;
         };
         GVAR(buildingModelCache) setVariable [_string, true];
     };
-} forEach (("configName _x isKindof 'building'" configClasses (configFile >> "CfgVehicles")) apply {getText (_x >> "model")});
+} forEach (("private _name = configName _x; _name isKindof 'building' || {_name isKindOf 'Rocks_base_F'}" configClasses (configFile >> "CfgVehicles")) apply {getText (_x >> "model")});
```
jokoho48, line 13: when we check here for Rocks_base_f we also should check that in shouldSpressPosition L37

diwako: why? the cache can be filled with any model that is supposed to block suppression, why not add more to it? All checks later will be then "does the name space have this key?"

diwako: That said the building check could also be dropped then

jokoho48: also i think the getText (_x >> "model") should be done in the loop it self. without we are looping 2x over all of those entries.

diwako: done.

addons/main/functions/fnc_shouldSuppressPosition.sqf
```diff
@@ -0,0 +1,44 @@
...
+ * [bob, angryJoe] call lambs_main_fnc_shouldSuppressPosition;
+ *
+ * Public: No
+*/
+params ["_unit", "_target"];
+
+#define NUM_PROBES 2
+
+if (_target isEqualType objNull) then {
+    _target = ATLtoASL (_target call CBA_fnc_getPos)
```
jokoho48, line 23: if you know it is an object why not just use getPosASL?

nk3nny (approved):

PR #254, opened 2021-05-13
Tweaks brainEngage.sqf for better assaults and suppression

nk3nny:
- Adds movement condition check for suppression to take place
- Tweaks unit assault to account for new indoor movement tweaks

### DIFF
--- a/addons/danger/functions/fnc_brainEngage.sqf
+++ b/addons/danger/functions/fnc_brainEngage.sqf
@@ -61,6 +61,11 @@ if (
     _timeout + 4
 };
 
+// running are busy and shouldn't suppress
+if ((speed _unit) isEqualTo 0) exitWith {
+    _timeout + 1
+};
+
 // set speed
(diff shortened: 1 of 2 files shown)


nk3nny:
Reverts to "_unit getHideFrom _target"  in doAssault.sqf

The change to actual position was done because in older FSM structures the assumed location would sometimes be [0,0,0]. I've not seen that recently.  More problematic. The assumed location is sometimes inside terrain features (or up the air). So this feature will need some watching. (and testing on nopryl!)

addons/danger/functions/fnc_brainEngage.sqf
```diff
@@ -61,6 +61,11 @@ if (
     _timeout + 4
 };
 
+// running are busy and shouldn't suppress
+if ((speed _unit) isEqualTo 0) exitWith {
```
nk3nny, line 65: Interesting. My intention was to run this as _isNotEqualTo_, yet this worked fine.  Will need a bit more testing :/

PR #256, opened 2021-05-13
Fix units attacking SuppressTargets

nk3nny:
LAMBS Danger.fsm adds more instances of the local vehicle "SuppressTarget". On occasion, units will chose them as valid 'ATTACK' targets.  This causes them to stupidly shoot at single locations.

Seems like a vanilla error where SuppressTargets are not always cleaned up.

This, crudely, fixes this by simply deleting suppressTargets which have turned into objects that are instead attacked.

### DIFF
--- a/addons/danger/functions/fnc_brainForced.sqf
+++ b/addons/danger/functions/fnc_brainForced.sqf
@@ -41,9 +41,11 @@ if (fleeing _unit) exitWith {
 
 // attack speed and stance
 if ((currentCommand _unit) isEqualTo "ATTACK") then {
-    _unit setVariable [QEGVAR(main,currentTask), "Attacking", EGVAR(main,debug_functions)];
-    [_unit, getAttackTarget _unit] call EFUNC(main,doAssaultSpeed);
+    private _attackTarget = getAttackTarget _unit;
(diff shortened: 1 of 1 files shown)


addons/danger/functions/fnc_brainForced.sqf
```diff
@@ -41,9 +41,11 @@ if (fleeing _unit) exitWith {
 
 // attack speed and stance
 if ((currentCommand _unit) isEqualTo "ATTACK") then {
-    _unit setVariable [QEGVAR(main,currentTask), "Attacking", EGVAR(main,debug_functions)];
-    [_unit, getAttackTarget _unit] call EFUNC(main,doAssaultSpeed);
+    private _attackTarget = getAttackTarget _unit;
+    if ((typeOf _attackTarget) isEqualTo "SuppressTarget") exitWith {deleteVehicle _attackTarget;};
```
diwako, line 45: unsure if that will cause issue along the way. At some point an ai will end up with `objNull` as target. I hope we have checks for that in place, else they will fire or look at [0,0,0]

nk3nny: Testing on nopryl server.

nk3nny: Mind you. In my experience, if an 'ATTACKING' unit has a objNull target, they drop out of the BIS attack fsm.

Issue #212, opened 2021-04-18
Re-implement Panic features ala 2.4.4


nk3nny:
Re-implement more extensive panic features as seen in older version of the mod. 

https://youtu.be/R8dwZ0bjKgk

zharf:
what's the status of getting this merged? we've been running a version where I cherry picked this for a few months now without any apparent issues

PR #260, opened 2021-05-22
Move panic feature from brain to eventhhandlers

nk3nny:
This pull request moves the panic feature from the brain to the eventhandler. This has the benefit of simplifying the FSM and making the panic feature more dependable.

The panic feature will not trigger for units that have a positive morale score, are not suppressed, or have their AI forced in some way. The panic feature is enhanced by reintroducing the panic types existant in lambs danger fsm 2.4.4

Answers to issue: https://github.com/nk3nny/LambsDanger/issues/212


jokoho48 (commented):
fix tabbing

PR #261, opened 2021-05-23
Make vehicle's main gun switch to HE munition when suppressing position

diwako:

1. Makes vehicles switch to HE munition in their main gun if they have that
2. Makes them switch back to their previous loaded magazine after suppression is done

The switching back is needed else tanks will always be shooting HE, also at other tanks.

- [x] Fix vehicles getting stuck on loaded ammunition type to be HE
- [x] Tanks suppress enemy tanks with HE now
- [x] Refactor the function some more

Right now this is a proof of concept, it needs more fleshing out.

### DIFF
--- a/addons/main/functions/VehicleAction/fnc_doSelectHEMunition.sqf
+++ b/addons/main/functions/VehicleAction/fnc_doSelectHEMunition.sqf
@@ -16,10 +16,9 @@
  * Public: No
 */
 params ["_vehicle"];
-private _pos = getPosATL player;
 private _gunner = gunner _vehicle;
 
-if (isNull _gunner) exitWith {};
+if (isNull _gunner) exitWith {false};
 
 // figure out turret and ammo
 private _turretPath = (assignedVehicleRole _gunner) select 1;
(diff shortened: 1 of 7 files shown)


addons/main/functions/VehicleAction/fnc_doSelectHEMunition.sqf
```diff
@@ -0,0 +1,99 @@
...
+            };
+        } forEach (if (_x == "this") then {
+            private _arr = _validMags arrayIntersect getArray ((configFile >> "CfgWeapons" >> _turret >> "magazines"));
+            reverse _arr;
+            _arr
+        } else {
+            private _arr = _validMags arrayIntersect getArray ((configFile >> "CfgWeapons" >> _turret >> _x >> "magazines"));
+            reverse _arr;
+            _arr
+        });
```
jokoho48, line 64: i dont like this code part here. i feel like this should be done before the if and than just use the _arr variable in the foreach.

addons/main/settings.sqf
```diff
@@ -50,6 +50,16 @@ private _curCat = LSTRING(Settings_MainCat);
     1
 ] call CBA_fnc_addSetting;
 
+// Toggles AI vehicle autonomous munition switching
+[
```
jokoho48, line 54: do we want to do this like the other systems and use "disable" instead of "enable"? most of the other systems have settings like that

jokoho48 (approved):

PR #263, opened 2021-06-04
Use Native Hashes for Artillery System

jokoho48:

1. *Title*


addons/wp/XEH_preInit.sqf
```diff
@@ -6,17 +6,17 @@ ADDON = false;
 if (isServer) then {
     [QGVAR(RegisterArtillery), {
         {
-            private _artillery = [GVAR(SideArtilleryHash), side _x] call CBA_fnc_hashGet;
+            private _artillery = GVAR(SideArtilleryHash) get (side _x);
```
diwako, line 9: For safety i would suggest `getOrDefault` instead of a simple get.

jokoho48: Yeah true and default should be a empty array 👍

diwako (approved):

PR #264, opened 2021-06-04
Fix Issues that Artillery Guns not Getting Readded in certain cases.

jokoho48:

1. *Title*


addons/wp/functions/fnc_doArtillery.sqf
```diff
@@ -163,7 +169,9 @@ if (_markerList isNotEqualTo []) then {
...
 
 // re-add to list
-if (!canFire _gun) exitWith {false};
+if (!canFire _gun) exitWith {
+    [QGVAR(RegisterArtillery), [_gun]] call CBA_fnc_serverEvent;
+    false
+};
 
 // ready up again
-[QGVAR(RegisterArtillery), [_gun]] call CBA_fnc_serverEvent;
```
diwako, line 169: Why remove that? Artillery gun will then not be added if they successfully fired and are ready to fire more?

jokoho48: Because I am stupid xD will revert that

diwako (approved):

PR #272, opened 2021-06-20
Concept for configurable range settings

jokoho48:

1. *Title*


diwako:
Not really a fan that everything needs to be packed now.

PR #273, opened 2021-06-21
Wrap all brain calls into a directCall

jokoho48:

1. *Title*

This is just a Idea that we may want to test out and see how it affects performance


diwako:
Imo needs extensive testing, not sure if that will cause any issues or not.

rekterakathom:
Would this even do anything? As far as I can tell, all of LAMBS is already in unscheduled space. The wiki doesn't explicitly say that code called from an FSM runs unscheduled (it also doesn't say that it runs scheduled), but all my tests point towards unscheduled execution.

For example, all brain calls are clearly visible in Arma Script Profiler, which only shows unscheduled functions. ``diag_activeSQFScripts`` also doesn't contain anything LAMBS related.

So given that we are already in unscheduled space, wouldn't this just add a tiny bit of overhead to everything?

PR #274, opened 2021-06-23
Fix the 3den Artillery Register Checkbox was not working

jokoho48:

1. *Title*

### DIFF
--- a/addons/wp/Cfg3DEN.hpp
+++ b/addons/wp/Cfg3DEN.hpp
@@ -37,7 +37,7 @@ class Cfg3DEN {
                         control = "Checkbox";
                         displayName = CSTRING(3DEN_Attributes_RegisterArtillery_DisplayName);
                         tooltip = CSTRING(3DEN_Attributes_RegisterArtillery_ToolTip);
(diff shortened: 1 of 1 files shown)


addons/wp/Cfg3DEN.hpp
```diff
@@ -37,7 +37,7 @@ class Cfg3DEN {
                         control = "Checkbox";
                         displayName = CSTRING(3DEN_Attributes_RegisterArtillery_DisplayName);
                         tooltip = CSTRING(3DEN_Attributes_RegisterArtillery_ToolTip);
-                        expression = "if (_value) then {_this spawn {waitUntil {!isNil 'lambs_danger_Loaded_WP'};if (lambs_danger_Loaded_WP) then {[gunner _this] call lambs_wp_fnc_taskArtilleryRegister;};};};";
```
diwako, line 40:
```sqf
if (_value) then {
    [] spawn {
        waitUntil {!(isNil 'lambs_wp_fnc_taskArtilleryRegister')};
        [gunner _this] call lambs_wp_fnc_taskArtilleryRegister;
    };
};
```
Why is [] passed to the spawn and not `_this`?

PR #276, opened 2021-06-29
Move min obstacle proximity to setting

jokoho48:

1. *Title*


diwako (changes requested):
Remove the line here, else it gets overwritten on mission start

https://github.com/nk3nny/LambsDanger/blob/01edf1aa0ca7c64b335eed35da90ffe67ac6c14c/addons/main/XEH_preInit.sqf#L14

diwako (approved):

nk3nny (approved):

PR #277, opened 2021-07-07
Polish translation - update for LAMBS 2.5.2 Dev 2

MiszczuZPolski:

1. Add missing stringtables
2. I changed name of one module to avoid misunderstaning, even if they're similar (Szturm->Atak in "LAMBS Danger Waypoints")
3. "Punkty trasy" was changed to "Punkty nawigacyjne" because that's name for them in vanilla


addons/wp/stringtable.xml
```diff
@@ -679,12 +684,12 @@
         </Key>
         <Key ID="STR_Lambs_WP_Context_Main_DisplayName">
             <English>Lamber Danger Waypoints</English>
-            <Polish>LAMBS Danger Punkty trasy</Polish>
+            <Polish>Lamber Danger Waypoints</Polish>
```
jokoho48, line 687: LAMBS is actually correct here :P looks like there is a misspelling in the English version in this case.

MiszczuZPolski: Because I coulnd't find appropriate name for difference between "LAMBS Danger Waypoints" and "LAMBS Danger Search Waypoints", I copy-pasted them from english (tbh I didn't see that "Lamber")

jokoho48:
@veteran29 could you plox do a review THANK YOU

t-zilla (approved):

addons/wp/stringtable.xml
```diff
@@ -375,7 +376,7 @@
         <Key ID="STR_Lambs_WP_Module_TaskPatrol_Range_Tooltip">
             <English>Max distance between waypoints</English>
             <German>Maximale Distanz zwischen zwei Wegpunkten.</German>
-            <Polish>Maksymalny zasięg pomiędzy punktami trasy</Polish>
+            <Polish>Maksymalny zasięg pomiędzy punktami nawigacyjnymi</Polish>
```
veteran29, line 379:
```suggestion
            <Polish>Maksymalny dystans pomiędzy punktami nawigacyjnymi</Polish>
```

Wydaje mi się że "zasięg" nie pasuje w tym kontekście.

addons/wp/stringtable.xml
```diff
@@ -693,64 +698,83 @@
...
-            <Polish>Jednostka będzie odpowiadać na zagrożenia wykryte przez sojuszników. \nWażne: Ta opcja będzie nadpisywać istniejące punkty trasy grup odpowiadających na wezwanie.</Polish>
+            <Polish>Jednostka będzie odpowiadać na zagrożenia wykryte przez sojuszników. \nWażne: Ta opcja będzie nadpisywać istniejące punkty nawigacyjne grup odpowiadających na wezwanie.</Polish>
         </Key>
         <Key ID="STR_Lambs_WP_Waypoints_displayName">
             <English>Advanced AI [LAMBS]</English>
+            <Polish>Advanced AI [LAMBS]</Polish>
         </Key>
         <Key ID="STR_Lambs_WP_Waypoint_TaskAssault_displayName">
             <English>TASK ASSAULT</English>
+            <Polish>Zadanie - Atak</Polish>
```
veteran29, line 709: Not sure if these should not be left upper case.

@jokoho48 ?

veteran29 (changes requested):

addons/wp/stringtable.xml
MiszczuZPolski: Tak, dystans wydaje się lepszym rozwiązaniem, niedługo to poprawię.

addons/wp/stringtable.xml
diwako: Should be upper case imo

MiszczuZPolski: I don't think if that makes a difference, but I can change it.

veteran29 (approved):

PR #278, opened 2021-07-19
use Native Hashes in doCallout, showDialog and shouldSuppressPosition

jokoho48:

1. *use Native Hashes in doCallout, showDialog and shouldSuppressPosition*
2. *mini cleanup in parseData from ShowDialog*

### DIFF
--- a/addons/main/functions/fnc_shouldSuppressPosition.sqf
+++ b/addons/main/functions/fnc_shouldSuppressPosition.sqf
@@ -37,7 +37,7 @@ private _fnc_checkIfBlocked = {
     private _index = _vis findIf {
         private _obj = (_x select 2);
         isNull _obj || // terrain geom
-        {GVAR(blockSuppressionModelCache) getVariable [((getModelInfo _obj) select 1), false]} // object that has a model that blocks suppression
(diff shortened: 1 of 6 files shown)


addons/main/functions/fnc_shouldSuppressPosition.sqf
```diff
@@ -37,7 +37,7 @@ private _fnc_checkIfBlocked = {
     private _index = _vis findIf {
         private _obj = (_x select 2);
         isNull _obj || // terrain geom
-        {GVAR(blockSuppressionModelCache) getVariable [((getModelInfo _obj) select 1), false]} // object that has a model that blocks suppression
+        {GVAR(blockSuppressionModelCache) getOrDefault [((getModelInfo _obj) select 1), false]} // object that has a model that blocks suppression
```
diwako, line 40: Native hashes are case sensitive, you are breaking the suppression feature here. You'd need to lower case the model info path here. The reason why a namespace was used here was to have case agnostic getter.

The model path can have any case in config, the keys inside the hashmap are all lower case.

diwako (changes requested):

jokoho48:
after thinking about this, it is probably not reasonable to use native hashes in those places

PR #284, opened 2021-09-27
implement smoke grenade usage as a config option

Couls:

1. Re-add the smoke grenade usage config option
2. Prevent AI from using smoke grenades if the CBA option is checked

I put the check before doSmoke is called, unsure if it would be better to just do one check after doSmoke is called versus preventing doSmoke from being called if the check doesn't pass, easy to change regardless so let me know which one would be better

### DIFF
--- a/addons/danger/functions/fnc_tacticsAssault.sqf
+++ b/addons/danger/functions/fnc_tacticsAssault.sqf
@@ -77,7 +77,7 @@ _group enableAttack false;
 [_unit, "combat", "Advance", 125] call EFUNC(main,doCallout);
 
 // concealment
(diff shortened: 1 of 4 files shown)


nk3nny:
This is looking very nice @Couls. Two small things: 
- Prefer to use lazy evaluation for the if clauses. 
- Double check that the strings and their translations are in place! 

:)

Couls:
> * Double check that the strings and their translations are in place!
they're already in place on the main repo here https://github.com/nk3nny/LambsDanger/blob/3abc64246b44f7c91fede4628c65a676f930b141/addons/danger/stringtable.xml#L172

PR #304, opened 2021-12-15
Fix groups never suppressing as a group

nk3nny:
Tweaks the early suppression check-- adding height to ensure more reliable suppressing events.
Fixes missing enableAttack use

### DIFF
--- a/addons/danger/functions/fnc_tacticsSuppress.sqf
+++ b/addons/danger/functions/fnc_tacticsSuppress.sqf
@@ -23,7 +23,7 @@ params ["_unit", "_target", ["_units", []], ["_delay", 17]];
 _target = _target call CBA_fnc_getPos;
 
 // exit with flank squad leader cannot suppress from here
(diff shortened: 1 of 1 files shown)


jokoho48 (approved):

addons/danger/functions/fnc_tacticsSuppress.sqf
```diff
@@ -65,8 +65,9 @@ _pos pushBack _target;
 _unit setVariable [QEGVAR(main,currentTarget), _target, EGVAR(main,debug_functions)];
 _unit setVariable [QEGVAR(main,currentTask), "Leader Suppress", EGVAR(main,debug_functions)];
 
-private _group = group _unit;
 // set group task
+private _group = group _unit;
+_group enableAttack false;
```
diwako, line 70: when will this be set to true again?

nk3nny: L39 of same file.  When the reset is configured.  Actually toggling attack mode was missing.

PR #306, opened 2021-12-15
Improve assault memory and engagement routines

nk3nny:
This pull request tunes AI close quarter battle ability and introduces small bug fixes.  It does this by tuning behaviours and introducing a change in how groupMemory entries are handled by groups. _See below_.

**brainEngage.sqf**
- Tweaks various brainEngage timings for increased aggression
- Increased chance of suppressing enemyDetected (taking advantage of improved suppress AI added in previous releases)
- Fix Units will only set "ACE_medical_ai_lastFired" when in CQB
- Fix Units will no longer suppress quickly moving vehicles, such as aircraft

**brainAssess.sqf**
- Removes many features from brainAssess which were necessary with the earllier slower brainEngage timings.
- Slimmer and better performing.

**doAssault.sqf**
- Better handling of stance in doAssault. (More standing means more reliable movement)
- Increased building check distance to get more positions. From 5 to 7 meters.
- Improved groupMemory*
- Reintroduce slight randomness to indoor assault locations (prevents units getting stuck)

**doAssaultMemory.sqf**
- Adds distance sorting to nearby positions
- Adds 'break out' routine if threatened by known, nearby enemy.
- Added stance change if nearest location is beyond 66M
- changed from doLook to doWatch for better group focus.
- Changed to Stance tweak and removed speed change to remove staccato movement.

_Improved group Memory_
*Building positions are now only added to groupMemory if the list is currently empty and the calling unit is fairly close (40 meters) to the building in question. This enhancement makes groups more reliable clear building by building-- and prevents the groupMemory from becoming endlessly growing.

The groupMemory feature now also has breakout patterns if the AI is threatened by nearby enemies.  And finally it interacts very well with the tightened timings-- adding an aggressive and constantly searching AI.

### DIFF
--- a/addons/danger/functions/fnc_brainEngage.sqf
+++ b/addons/danger/functions/fnc_brainEngage.sqf
@@ -29,13 +29,11 @@ params ["_unit", ["_type", -1], ["_target", objNull]];
 // timeout
 private _timeout = time + 0.5;
 
-// ACE3
-_unit setVariable ["ace_medical_ai_lastFired", CBA_missionTime];
-
 // check
 if (
     isNull _target
     || {_unit knowsAbout _target isEqualTo 0}
+    || {(speed _target) > 20}
     || {(weapons _unit) isEqualTo []}
     || {(combatMode _unit) in ["BLUE", "GREEN"]}
 ) exitWith {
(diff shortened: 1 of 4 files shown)


addons/danger/functions/fnc_brainEngage.sqf
```diff
@@ -29,13 +29,11 @@ params ["_unit", ["_type", -1], ["_target", objNull]];
...
 private _timeout = time + 0.5;
 
-// ACE3
-_unit setVariable ["ace_medical_ai_lastFired", CBA_missionTime];
-
 // check
 if (
     isNull _target
     || {_unit knowsAbout _target isEqualTo 0}
+    || {(speed _target) > 20}
```
diwako, line 36: why 20? iirc running is like a speed of 11

nk3nny: Arbitrary number to be honest.  I observed AI flying slowly in a few different helicopters.  Wanted to pick something where they would suppress when moving slowly, but not when moving too quickly.

diwako (approved):

PR #308, opened 2021-12-15
Improve Assault group tactics

nk3nny:
This pull request adds a significantly improved assault pattern.  At its core the biggest change are tuned timings and distances, and introducing sorting of building/enemy locations by distance to leader.  This means that the squad will converge on each location in an organised fashion-- where previously they did so more randomly in a shotgun pattern.

_doGroupAssault.sqf_ has been improved with breakout patterns (when an AI is immediately threatened by AI) and a better tuned movement speed-- meaning the units will move with their weapons raised, rather than sprinting. Obviously more conducive to assaulting...

### DIFF
--- a/addons/main/functions/GroupAction/fnc_doGroupAssault.sqf
+++ b/addons/main/functions/GroupAction/fnc_doGroupAssault.sqf
@@ -31,8 +31,8 @@ private _targetPos = _pos deleteAt 0;
     _x setVariable [QEGVAR(danger,forceMove), true];
 
     // check enemy
-    private _enemy = _unit findNearestEnemy _unit;
-    if (_unit distance2D _enemy < 12) then {_targetPos = getPosATL _enemy;};
+    private _enemy = _x findNearestEnemy _x;
+    if (_x distance2D _enemy < 12) then {_targetPos = getPosATL _enemy;};
 
     // setpos
     if (RND(0.75) || {(currentCommand _x) isNotEqualTo "MOVE"}) then {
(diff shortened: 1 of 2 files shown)


addons/main/functions/GroupAction/fnc_doGroupAssault.sqf
```diff
@@ -20,30 +20,34 @@ params ["_cycle", "_units", "_pos"];
...
-
     // manoeuvre
-    _x forceSpeed 4;
+    _x forceSpeed 3;
     _x setUnitPosWeak (["UP", "MIDDLE"] select (getSuppression _x isNotEqualTo 0));
     _x setVariable [QGVAR(currentTask), "Group Assault", GVAR(debug_functions)];
     _x setVariable [QEGVAR(danger,forceMove), true];
+
+    // check enemy
+    private _enemy = _x findNearestEnemy _x;
```
diwako, line 34: ne need to assign a variable

nk3nny: Would 
`
if (_x distance2D (_x findNearestEnemy _x) < 12) then {_targetPos = getPosATL (_x findNearestEnemy _x);};
`
Be better.  Or am I misunderstanding something?

diwako: it is readability vs speed in this case.

I would prefer speed here as assigning a variable that is only used once seems a bit of a waste to me.

nk3nny: kk

nk3nny: Fixed

PR #309, opened 2021-12-16
Improved tactics assess

nk3nny:
- Adds sorting enemies based on distance
- Fixes groups failing to deploy static weapons (and sometimes UGL flares)
- Changed - reverts Suppression chance to accomodate #304
- Removed vanilla attack feature from groups*

*This is a big one which we've talked about internally before. What this does is remove the lock-in units sometimes suffer due to the vanilla ATTACK pattern/fsm.  I propose to check this on the development version.

Our own _group_ level attack patterns are better than the individual vanilla ones. But the AI is a complicated beast. Additional testing should be done.

### DIFF
--- a/addons/danger/functions/fnc_tacticsAssess.sqf
+++ b/addons/danger/functions/fnc_tacticsAssess.sqf
@@ -33,6 +33,7 @@ private _group = group _unit;
 // set variable
 _group setVariable [QGVAR(isExecutingTactic), true];
 _group setVariable [QGVAR(contact), time + 600];
+_group enableAttack false;
 
 // set current task
 _unit setVariable [QEGVAR(main,currentTarget), objNull, EGVAR(main,debug_functions)];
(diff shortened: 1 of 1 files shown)


addons/danger/functions/fnc_tacticsAssess.sqf
```diff
@@ -33,6 +33,7 @@ private _group = group _unit;
 // set variable
 _group setVariable [QGVAR(isExecutingTactic), true];
 _group setVariable [QGVAR(contact), time + 600];
+_group enableAttack false;
```
diwako, line 36: when will this set to true again? Any contingencies if a mission maker wants to set it themselves?

nk3nny: At this point there is none.  As noted in the pull description it is something I'd like to run larger scale testing in on a development version. 

1 I guess the code should reflect that with more aggressive commenting (will fix). 
2. Future versions could include a CBA setting. Though, if people want to _enjoy_ the vanilla attack patterns, they can simply disable LAMBS group AI -- or group manoeuvres alltogether.

We've played with these settings on nopryl for the past months. Observed experiences are: 
- No negative issues
- Fewer instances of AI going off on individual suicide missions
- Fewer instances of AI being drawn to attack from otherwise defensive positions
- Overall, better squad cohesion.

PR #317, opened 2022-04-18
Improvements to Engagement

nk3nny:
**Yeah. There is a typo in the branch name.** 

**Changes**
Adds groupMemory check when CQB engaging
Adds visibilty check to determine rooms secured
Adds fix to attempting to suppress targets above ground
Improves FSM by making dodging trigger a "can-fire" rather than assessment state.
Fixes use of lookAt to doWatch

**Description**
Generally improves use of lookAt which fixes AI getting information they are not supposed to have.

Better, more consistent handling og groupMemory, meaning that AI will more effectively clear buildings as teams. Also tweaks many settings to ensure infantry move more intelligently to check buildings. Including visual checks to determine if building positions are cleared.  This fixes units 'clearing' buildings from the outside.

### DIFF
--- a/addons/danger/functions/fnc_brainAssess.sqf
+++ b/addons/danger/functions/fnc_brainAssess.sqf
@@ -26,7 +26,7 @@
 params ["_unit", "", "", ["_target", objNull]];
 
 // timeout
-private _timeout = time + 2;
+private _timeout = time + 4;
 
 // check if stopped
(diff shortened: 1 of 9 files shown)


diwako (approved):
Looks good. just a fair warining that `checkVisibility` is really slow and should not be used _too_ much.

jokoho48 (approved):

PR #320, opened 2022-04-20
Fix params type for the position argument in taskPatrol

Keithenneu:

1. Fix the params type for the position argument in taskPatrol. It now uses the same type (any) as taskGarrison

The limited type would show script errors when using objects or groups, and then default to the group position. Which probably was the desired position in most cases anyways tbf.


diwako (approved):
LGTM

`_pos` is handled by the CBA function anyways.

jokoho48 (approved):

PR #329, opened 2022-12-18
Improve assault behaviour

nk3nny:
Added reload check before group assaults
Added debug markers to various Contact states
Added visibility checks to many clearing routines
Added prioritisation of enemies on the same floor to many clearing routines

Improved doAssault and doAssaultMemory
Improved cqb reaction speed by removing doWatch from many functions Improved tacticsAssault building selection criteria Improved FSM exit condition when the group has memory

Fixed unecessary FSM reset when AI is forced
Fixed tacticsGarrison rarely functioning


nk3nny:
<3 linter

PR #331, opened 2023-02-19
Add Error when Loading Lambs only on Clients but not on the Server

jokoho48:

1. *Title*

### DIFF
--- a/addons/main/XEH_preInit.sqf
+++ b/addons/main/XEH_preInit.sqf
@@ -16,6 +16,7 @@ if (isServer) then {
     publicVariable QGVAR(versionLoadedOnServer);
 } else {
     0 spawn {
+        waitUntil {time > 0};
         if (!isNil QGVAR(versionLoadedOnServer) && GVAR(versionLoadedOnServer) isEqualTo QUOTE(VERSION_STR)) exitWith {};
         "LAMBS DANGER VERSION MISMATCH ERROR!!!!" hintC parseText format [
             "Lambs Danger Version mismatch Error.<br/>Client Version: %1<br/>Server Version: %2",
--- a/addons/main/XEH_preInit.sqf
+++ b/addons/main/XEH_preInit.sqf
@@ -16,7 +16,7 @@ if (isServer) then {
     publicVariable QGVAR(versionLoadedOnServer);
 } else {
     0 spawn {
-        if (GVAR(versionLoadedOnServer) isEqualTo QUOTE(VERSION_STR)) exitWith {};
+        if (!isNil QGVAR(versionLoadedOnServer) && GVAR(versionLoadedOnServer) isEqualTo QUOTE(VERSION_STR)) exitWith {};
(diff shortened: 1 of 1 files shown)


addons/main/XEH_preInit.sqf
```diff
@@ -11,4 +11,29 @@ GVAR(shareHandlers) = [];
 
 GVAR(blockSuppressionModelCache) = false call CBA_fnc_createNamespace;
 
+if (isServer) then {
+    GVAR(versionLoadedOnServer) = QUOTE(VERSION_STR);
+    publicVariable QGVAR(versionLoadedOnServer);
+} else {
+    0 spawn {
+        if (GVAR(versionLoadedOnServer) isEqualTo QUOTE(VERSION_STR)) exitWith {};
```
diwako, line 19: Nil check for `GVAR(versionLoadedOnServer)` after mission start.

diwako (changes requested):

addons/main/XEH_preInit.sqf
```diff
@@ -11,4 +11,29 @@ GVAR(shareHandlers) = [];
...
+            "Lambs Danger Version mismatch Error.<br/>Client Version: %1<br/>Server Version: %2",
+            QUOTE(VERSION_STR),
+            GVAR(versionLoadedOnServer)
+        ];
+        private _error = format [
+            "Lambs Danger Version mismatch Error. Client Version: %1 Server Version: %2",
+            QUOTE(VERSION_STR),
+            GVAR(versionLoadedOnServer)
+        ];
+        while {true} do {
```
diwako, line 30: you sure you want to spam that every second?

jokoho48: yes

addons/main/XEH_preInit.sqf
```diff
@@ -11,4 +11,30 @@ GVAR(shareHandlers) = [];
 
 GVAR(blockSuppressionModelCache) = false call CBA_fnc_createNamespace;
 
+if (isServer) then {
+    GVAR(versionLoadedOnServer) = QUOTE(VERSION_STR);
+    publicVariable QGVAR(versionLoadedOnServer);
+} else {
+    0 spawn {
+        waitUntil {time > 0};
+        if (!isNil QGVAR(versionLoadedOnServer) && GVAR(versionLoadedOnServer) isEqualTo QUOTE(VERSION_STR)) exitWith {};
```
veteran29, line 20: Needs lazy eval, otherwise it's still comparing nil to string

diwako: I blame Joko

jokoho48: fixed on master ^^ thanks. but stupid SQF 🤣

rautamiekka: > stupid SQF

Amen to that. _Won't be getting much better ..._

PR #333, opened 2023-02-23
Adds better support for MLRS artillery

nk3nny:
**Improves Dynamic Artillery support for MLRS style launchers.** 
- MLRS launchers will now shoot their entire magazine. Launchers with large magazines will still clump their salvos in 2's or 4's to ensure good spread on target. 
- MLRS launchers will have no check rounds-- which make them quite dangerous in numbers.

### DIFF
--- a/addons/wp/functions/fnc_doArtillery.sqf
+++ b/addons/wp/functions/fnc_doArtillery.sqf
@@ -72,7 +72,7 @@ if (_isMLRS isEqualTo -1) then {
     // get the callout for what this vehicle shoots!
     private _turretPath = _assignedRoles select 1;
     private _turret = (_gun weaponsTurret _turretPath) select 0;
-    _nameSound = getText (configFile >> "CfgWeapons" >> _turret >> "nameSound");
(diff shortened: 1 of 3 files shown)


rautamiekka:
I suppose the idea is to make sure the dynamic target, and everything around it, is gonna die ?

Or is this a bug fix of sorts ?

diwako:
> I suppose the idea is to make sure the dynamic target, and everything around it, is gonna die ?
> 
> Or is this a bug fix of sorts ?

It is a lowkey feature request from someone from our discord server. They are developing MRLS artillery vehicle assets and asked if it was intended for them to shoot only one rocket at a time.
The current code more or less only simulates tube based artillery and is tweaked for gameplay. With this you can mark an artillery as MRLS, so it behaves a bit different and actually uses the rocket/missile salvo as an MRLS would do. Pretty much obliterating a whole area as their original intend is.

rautamiekka:
Interesting, why not.

nk3nny:
Remember to add Zeus interaction

jokoho48 (changes requested):
Zeus Modules and Editor Modules missing. also TaskArtillery has header not updated

PR #340, opened 2023-03-03
Utilize insideBuilding

rekterakathom:
**Title**
Utilize the new[ ``insideBuilding``](https://community.bistudio.com/wiki/insideBuilding) command introduced in 2.12. In my tests it is ~10x faster than the old method. Also way more reliable since this is an engine solution. Requires 2.12 though!
3. *Each change in a separate line*
Changed raycast to ``insideBuilding``

### DIFF
--- a/addons/main/functions/fnc_isIndoor.sqf
+++ b/addons/main/functions/fnc_isIndoor.sqf
@@ -15,6 +15,4 @@
  * Public: No
 */
 params ["_unit"];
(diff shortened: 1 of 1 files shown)


nk3nny:
Thanks!

nk3nny:
I've tested this quite extensively, and the _insideBuilding_ check (though faster) fails to capture many buildings.  I.e., any building with an open wall.

nk3nny:
Further testing with combining the older system (and the newer one) unfortunately is overall slower-- as we check much more often outside than inside. :)

PR #341, opened 2023-03-03
Rework position commands

rekterakathom:
Replace all instances of ``getPos`` with the appropriate position command, usually ``ASLtoAGL getPosASL``. This is because ``getPos`` is not a suitable command for getting an objects 3D position due to [inconsistent behaviour and slow execution](https://community.bistudio.com/wiki/getPos).

Add a new macro for getting 3D positions in positionAGL format and utilize it to replace all instances of ``getPos``
3. *Each change in a separate line*
New macro ``POSITIONAGL(object)``
Removed ``getPos`` in favour of the new macro

### DIFF
--- a/addons/danger/functions/fnc_tacticsAssess.sqf
+++ b/addons/danger/functions/fnc_tacticsAssess.sqf
@@ -104,7 +104,7 @@ if !(_enemies isEqualTo [] || {_unitCount < random 4}) then {
     if (EGVAR(main,Loaded_WP) && {[side _unit] call EFUNC(WP,sideHasArtillery)}) then {
         private _artilleryTarget = _enemies findIf {
             _unit distance2D _x > RANGE_MID
-            && {([_unit, getPos _x, RANGE_NEAR] call EFUNC(main,findNearbyFriendlies)) isEqualTo []}
+            && {([_unit, POSITIONAGL(_x), RANGE_NEAR] call EFUNC(main,findNearbyFriendlies)) isEqualTo []}
         };
         if (_artilleryTarget != -1) then {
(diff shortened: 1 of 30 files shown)


nk3nny:
Nice PR. 

@diwako 
For more discussion.

nk3nny:
I like this solution. It seems to run significantly faster in most situations, with the odd exceptions of terrain objects gathered from  _nearestTerrainObjects_, which is used in some hiding and fleeing functions. 

https://community.bistudio.com/wiki/nearestTerrainObjects

nk3nny:
@diwako 
Let's have a discussion once you are back about this one. :)

nk3nny:
Allright. Review completed.  Only to half a year. 

We like the concept and appreciate the thought that has gone into it.  However we are suspect about the maintainability (see review time) and would like instead to implement non-macroed version in key positions. :)

rekterakathom:
> Allright. Review completed. Only to half a year.
> 
> We like the concept and appreciate the thought that has gone into it. However we are suspect about the maintainability (see review time) and would like instead to implement non-macroed version in key positions. :)

The way I see it, partially implementing this PR w/o macros is the less maintainable version. That way you'll end up with various position commands across the codebase, whereas with this PR you get universal getters / setters for the most common position format with singular definitions.

rautamiekka:
> The way I see it, partially implementing this PR w/o macros is the less maintainable version. That way you'll end up with various position commands across the codebase, whereas with this PR you get universal getters / setters for the most common position format with singular definitions.

Yeah, hard to argue against the complexity that those macros are simplifying.

PR #345, opened 2023-04-16
Improve Group and Contact Suppression

nk3nny:
Added a follow-up fire state to suppression states
Improves interaction with _doAssaultMemory_ (more predictable) Fixes vehicles being part of Contact state

### DIFF
--- a/addons/danger/functions/fnc_tacticsContact.sqf
+++ b/addons/danger/functions/fnc_tacticsContact.sqf
@@ -127,15 +127,14 @@ if (
         [
             {
                 params ["_unit", "_posASL"];
-                if (_unit call EFUNC(main,isAlive)) then {
+                systemchat format ["%1 suppress follow-up with %2", name _unit, currentCommand _unit];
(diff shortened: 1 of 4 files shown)


addons/danger/functions/fnc_tacticsContact.sqf
```diff
@@ -155,13 +166,24 @@ if (
...
+        _x forceSpeed 3;
+        //_x doMove selectRandom _buildings;
         _x setVariable [QEGVAR(main,currentTask), "Assault (contact)", EGVAR(main,debug_functions)];
 
         // forced movement
         _x setVariable [QGVAR(forceMove), true];
-        [{_this setVariable [QGVAR(forceMove), nil]}, _x, 5 + random 4] call CBA_fnc_waitAndExecute;
-    } foreach _units select {getSuppression _x < 0.5};
+        [
+            {
```
diwako, line 178: Check if unit is dead and restore old unitPos

nk3nny: Good call. Fixed in latest push.

PR #350, opened 2023-04-17
Hemtt upgrade

diwako:
Upgrade to latest helm version and rework

- [x] Convert project file
- [x] Auto sign pbos
- [x] Auto compile SQFC
- [x] Set version in version.hpp
- [x] Set version in mod.cpp
- [x] Auto attach built version to releases


.github/workflows/main.yml
```diff
@@ -35,7 +35,7 @@ jobs:
     - name: Lint (sqflint)
       uses: jokoho48/sqflint@master
       continue-on-error: true # No failure due to many false-positives
-    - uses: actions/upload-artifact@master
+    - uses: actions/upload-artifact@v4.3.1
```
diwako, line 38: why not just use `v4` as that should be use the latest v4 version

.github/workflows/main.yml
```diff
@@ -48,14 +48,19 @@ jobs:
       - uses: actions/checkout@master
         with:
           fetch-depth: 1
-      - name: Build addon with HEMTT
-        uses: arma-actions/hemtt@master
-        with:
-          command: 'build --release'
-      - uses: actions/upload-artifact@master
+      - name: Setup Node.js environment
+        uses: actions/setup-node@v4.0.2
```
diwako, line 52: why not just `v4` ?

PR #354, opened 2023-04-29
Update fnc_doCheckBody.sqf

nk3nny:
Improves how the AI will check bodies:
- Adds the ability for AI to pick up backpacks of friendly units 
- Adds ability for AI to pick up AT/AA weapons from friendly units

Simply improves sustainability of squads in combat.

### DIFF
--- a/addons/main/functions/UnitAction/fnc_doCheckBody.sqf
+++ b/addons/main/functions/UnitAction/fnc_doCheckBody.sqf
@@ -50,7 +50,7 @@ _unit setVariable [QEGVAR(danger,forceMove), true];
             if ((backpack _unit) isEqualTo "" && {backpack _body isNotEqualTo ""}) then {
                 private _items = backpackItems _body;
                 private _backpack = backpack _body;
-                removeBackpack _body;
+                removeBackpackGlobal _body;
                 _unit addBackpack _backpack;
                 {_unit addItemToBackpack _x} forEach _items;
             };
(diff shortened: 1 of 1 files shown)


addons/main/functions/UnitAction/fnc_doCheckBody.sqf
```diff
@@ -37,14 +37,34 @@ _unit setVariable [QEGVAR(danger,forceMove), true];
...
         if (_unit call FUNC(isAlive)) then {
             [QGVAR(OnCheckBody), [_unit, group _unit, _body]] call FUNC(eventCallback);
             _unit action ["rearm", _body];
+
+            // get backpack
+            if ((backpack _unit) isEqualTo "" && {backpack _body isNotEqualTo ""}) then {
+                private _items = backpackItems _body;
+                private _backpack = backpack _body;
+                removeBackpack _body;
+                _unit addBackpack _backpack;
```
rautamiekka, line 54: Hmm, shouldn't these be the global variants ?

^ [The latter](https://community.bistudio.com/wiki/addBackpackGlobal)'s docs warn you need to `waitUntil` cuz the backpack will be null for a bit on a non-local unit, though, but I don't see why that would be a problem.
```suggestion
                removeBackpackGlobal _body;
                _unit addBackpackGlobal _backpack;
```

nk3nny: Nice change. Because we will only be adding backpacks to LOCAL units, we don't need to worry about waiting.

PR #356, opened 2023-07-02
Update isNight

rekterakathom:
### When merged this pull request will: Utilise ``getLighting`` to check if it's night

Utilise ``getLighting`` so the night check is based off of the amount of light instead of a soundcontroller. This might be an important distinction in cases where the night is really bright, due to midnight sun or the full moon.

~65 is the approximate light level where AI uses nightvision. Tested on Stratis (2035/07/06 YYYY/MM/DD) with vanilla AI.

- Switch ``getEnvSoundController`` to ``getLighting``

### DIFF
--- a/addons/main/functions/fnc_isNight.sqf
+++ b/addons/main/functions/fnc_isNight.sqf
@@ -23,4 +23,5 @@ if (behaviour _unit isEqualTo "STEALTH"
 ) exitWith {false};
 
 // night check
-(getPos _unit) getEnvSoundController "night" isEqualTo 1
+// ~65 is the light level where vanilla AI puts nightvision goggles on
+(getLighting # 1) < 65


nk3nny:
Hmm. Testing indicated that units on headless clients may have problems. HC reports different value for getLighting. 

It is possible to use _getLightingAt_, but that is actually slower than _getEnvSoundController_ .

https://community.bistudio.com/wiki/getLightingAt

rekterakathom:
Does it work on dedi servers though? If it does, then this definitely sounds like a HC bug.

> Hmm. Testing indicated that units on headless clients may have problems. HC reports different value for getLighting.
> 
> It is possible to use _getLightingAt_, but that is actually slower than _getEnvSoundController_ .
> 
> https://community.bistudio.com/wiki/getLightingAt

Does it work on dedi servers? If it does, then this sounds like a HC bug

nk3nny:
Tested on dedicated server environment. Both server and HC had  units spawned on it. 

I'll give it another test in a 'live' mission too.

nk3nny:
Server reported correct values. Only HC struggled. Wonder if there is some silliness about the location of the HC too.

diwako:
location about the hc unit should only affect update rates iirc. can always bother dedmen in the arma discord.

jokoho48:
> location about the hc unit should only affect update rates iirc. can always bother dedmen in the arma discord.

this is a "global" value getter not a local one. so I expect that there is some issue related to how HCs are handled.

rekterakathom:
FT tracker ticket: https://feedback.bistudio.com/T178396

PR #357, opened 2023-07-13
Optimize door finding in findBuildings

rekterakathom:
### When merged this pull request will: Optimize the door finding logic

Find doors faster by searching for the doors in ``viewGeometry`` ~~and using ``select`` to run the string filtering in-engine.~~

We can assume that doors are found in ``viewGeometry`` because this LOD is used for vision occlusion, which of course includes doors.

This new method is approximately ~4 times faster.

This also coincidentally fixes an issue where the same door positions get detected multiple times due to multiple selections having the same position, things like ``door_1`` and ``door_1_handle``.

``toLower`` is not necessary because all selections must be named lowercase.

Tested on random buildings on: Stratis, Livonia, Takistan, Chernarus, Lythium

**Further testing is encouraged!**

- Remove ``toLower`` ~~and separate condition for string filtering, replaced with in-engine loop using ``select`` that doesn't use ``toLower``~~ because we can assume that selections are named lowercase

- Search for selections in ``viewGeometry`` because we can assume that doors are found there

### DIFF
--- a/addons/main/functions/fnc_findBuildings.sqf
+++ b/addons/main/functions/fnc_findBuildings.sqf
@@ -37,8 +37,9 @@ private _housePos = [];
     _housePos append (_house buildingPos -1);
     if (_findDoors) then {
         {
+            if !("door" in _x) then {continue};
             _housePos pushBack (_house modelToWorld (_house selectionPosition _x));
-        } forEach ((_house selectionNames "viewgeometry") select {"door" in _x});
+        } forEach (_house selectionNames "viewgeometry");
     };
 } forEach _houses;
 
--- a/addons/main/functions/fnc_findBuildings.sqf
+++ b/addons/main/functions/fnc_findBuildings.sqf
@@ -37,8 +37,9 @@ private _housePos = [];
     _housePos append (_house buildingPos -1);
     if (_findDoors) then {
         {
-            if !("door" in _x) then {continue};
-            _housePos pushBack (_house modelToWorld (_house selectionPosition _x));
+            if ("door" in _x) then {
+                _housePos pushBack (_house modelToWorld (_house selectionPosition _x));
+            };
         } forEach (_house selectionNames "viewgeometry");
     };
 } forEach _houses;
--- a/addons/main/functions/fnc_findBuildings.sqf
+++ b/addons/main/functions/fnc_findBuildings.sqf
@@ -37,10 +37,8 @@ private _housePos = [];
     _housePos append (_house buildingPos -1);
     if (_findDoors) then {
         {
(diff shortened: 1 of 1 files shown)


addons/main/functions/fnc_findBuildings.sqf
```diff
@@ -37,10 +37,8 @@ private _housePos = [];
     _housePos append (_house buildingPos -1);
     if (_findDoors) then {
         {
-            if ("door" in toLower (_x)) then {
-                _housePos pushBack (_house modelToWorld (_house selectionPosition _x));
-            };
-        } forEach (selectionNames _house);
+            _housePos pushBack (_house modelToWorld (_house selectionPosition _x));
+        } forEach ((_house selectionNames "viewgeometry") select {"door" in _x});
```
jokoho48, line 41: I personally don't like that we loop 2x through the items here with first select and then forEach.  we should have the if still in the foreach

rekterakathom: Re-added the if. Select was faster, but this is definitely easier to read.

addons/main/functions/fnc_findBuildings.sqf
```diff
@@ -37,10 +37,9 @@ private _housePos = [];
     _housePos append (_house buildingPos -1);
     if (_findDoors) then {
         {
-            if ("door" in toLower (_x)) then {
-                _housePos pushBack (_house modelToWorld (_house selectionPosition _x));
-            };
-        } forEach (selectionNames _house);
+            if !("door" in _x) then {continue};
```
jokoho48, line 40: why do we use a continue here and additionally a negation. wouldn't it be just simpler to move the pushback into the if? like it was before.

rekterakathom: Because in my opinion less indentation is easier to read. Is there a specific reason you want it exactly like it was before?

jokoho48: in the end i expect negating and an extra script command maybe slower than that plus I personally dislike how it currently reads to be correctly formated it would occupy 3 lines and that just reads worse for me in total plus is unneeded longer
```sqf
if !("door" in _x) then {
    continue;
};
```

rekterakathom: Fair point with the speed, benchmarked the ``continue`` method to be ~0.05ms (rougly 5%) slower than the original.

But what do you mean correctly formatted? One line if statements are all over this project, some of them over a hundred characters long.

nk3nny:
Excellent stuff! @rekterakathom

PR #358, opened 2023-07-13
Improve explosionEH

### LINKED ISSUE #328: Issues with Webknights melee AI / Orks
**Mods (complete and add to the following information):**

- **Arma 3:** `3.15.2`
- **CBA:** `3.15.8`
- **LAMBS_DANGER:** `2.5.3`
- **[Greenskins](https://steamcommunity.com/sharedfiles/filedetails/?id=2599706429)** `latest, 7 dec 2022`

**Description:**

Not sure if this on LAMBS or webknights to fix, so I'll send this to both and hope it's possible to fix.

When a dead melee unit from Webknights Orks mod, for example a melee ork nob or a squig, gets hit by explosives the animation bugs out and the unit stands up again in a frozen state. (still dead)

This is "fixed" by disabling "Explosion Eventhandler" under "LAMBS Danger Eventhandlers" so there seems to be something there that tries to do something even on dead units. Again, I'm not sure in which addon this must be fixed, but hopefully you are able to look into it.

rekterakathom:
### When merged this pull request will: Improve explosionEH by adding more checks and using new features

Adds an alive check to the start, to make sure that code never gets run on dead units (hopefully fixes #328 )

Utilizes a new engine feature for getting the position of the explosion, allowing units to "flinch" accordingly

- Add alive and null checks to the start

- Utilize ``_source`` for getting a position to "flinch" from

### DIFF
--- a/addons/eventhandlers/functions/fnc_explosionEH.sqf
+++ b/addons/eventhandlers/functions/fnc_explosionEH.sqf
@@ -16,20 +16,22 @@
 */
 
 // init
-params ["_unit"];
+params [["_unit", objNull], ["_damage", 0], ["_source", objNull]];
 
 // Standing or recent explosions ignored
 if (
     !GVAR(ExplosionEventHandlerEnabled)
+    || isNull _unit
     || !(local _unit)
     || isPlayer _unit
     || {!isNull objectParent _unit}
     || {(stance _unit) isEqualTo "PRONE"}
     || {_unit getVariable [QGVAR(explosionReactionTime), 0] > time}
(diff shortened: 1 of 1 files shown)


addons/eventhandlers/functions/fnc_explosionEH.sqf
```diff
@@ -16,20 +16,22 @@
 */
 
 // init
-params ["_unit"];
+params [["_unit", objNull], ["_damage", 0], ["_source", objNull]];
```
jokoho48, line 19: as _damage is unused it is not required to be used here and should just be ""
```suggestion
params [["_unit", objNull], "", ["_source", objNull]];
```

PR #361, opened 2023-07-25
checkVisibility comparison fixes

rekterakathom:
### When merged this pull request will: Remove all instances of ``checkVisibility`` being compared against zero

- Remove all instances of checkVisibility being compared against zero. as [``checkVisibility`` will not return zero in case of vision blocking particles](https://community.bistudio.com/wiki/checkVisibility). With this PR, ``checkVisibility`` is compared against ``0.01`` instead. I have arbitrarily chosen 1% to be the visibility threshold, feel free to suggest otherwise.
- Remove all instances of checkVisibility being compared against zero.


diwako:
Thank you for the continued pull requests. Currently all authors are really busy with their real lives hence why things are slow

nk3nny:
Great stuff!

nk3nny (approved):

nk3nny:
This was a little delayed in getting approved simply because I wanted to play around about with the logic itself.  I think you are on to something on how building positions should be cleared. 

When testing this was done in a rather sterile environment, without any smoke disrupting LOS. I wonder if we want them to check extra carefully (not just disregard corners) as currently when there is smoke.  Regardless. This is good.

PR #364, opened 2023-08-20
Fixed type declaration

LordGolias:
This PR is a small fix to the type declaration of `params`.

When an array is passed to `params`, its third value is an array of objects representing their types (see example 5 in https://community.bistudio.com/wiki/params). This array was empty, which would represent that ~~no~~all types are allowed (see comment). Given that the default is an empty array, I suspect that what we mean is that parameter `_values` of `_fnc_AddDropDown` expects an object of type array. This PR does this.

Found while analyzing the code with https://marketplace.visualstudio.com/items?itemName=SQF-analyzer.sqf-analyzer

### DIFF
--- a/addons/main/functions/fnc_showDialog.sqf
+++ b/addons/main/functions/fnc_showDialog.sqf
@@ -157,7 +157,7 @@ private _fnc_AddBoolean = {
 };
 
 private _fnc_AddDropDown = {
-    params ["_text", "", ["_tooltip", ""], ["_values", [], []], ["_default", 0, [0]]];
+    params ["_text", "", ["_tooltip", ""], ["_values", [], [[]]], ["_default", 0, [0]]];
     if (isLocalized _tooltip) then {
         _tooltip = localize _tooltip;
     };


rekterakathom:
Not true, according to the biki, empty array means all data types are accepted.

``"expectedDataTypes: Array of direct Data Types - (Optional) checks if passed value is one of listed Data Types. If not, default value is used instead. Empty array [] means every data type is accepted."``

Doesn't make this PR useless though, as long as the data type restriction added here is correct. More parameter validation is (in my opinion) always ideal.

LordGolias:
indeed - sorry for providing the wrong information in the PR - fixed the PR.

jokoho48:
> indeed - sorry for providing the wrong information in the PR description - fixed. I believe the correction is still valid as this likely expects arrays (uses `forEach _values` and default is an empty array).

a foreach can handle more than just arrays: https://community.bistudio.com/wiki/forEach

LordGolias:
> > indeed - sorry for providing the wrong information in the PR description - fixed. I believe the correction is still valid as this likely expects arrays (uses `forEach _values` and default is an empty array).
> 
> a foreach can handle more than just arrays: https://community.bistudio.com/wiki/forEach

No-one said it otherwise.

Read the source code and you will find that
1. there is one invocation of this function
2. in this invocation, the argument is each entry of "_data" of the function `showDialog`
3. all invocations of the function `showDialog` in the entire code base pass an array to `_data`, and each element of the array is an array.

PR #366, opened 2023-09-10
Add Caching to Muzzle Check in doSmokeShell

jokoho48:

1. *Title*

### DIFF
--- a/addons/main/functions/fnc_getCompatibleThrowMuzzle.sqf
+++ b/addons/main/functions/fnc_getCompatibleThrowMuzzle.sqf
@@ -26,19 +26,17 @@ if (!isNil "_muzzle") exitWith {
     _muzzle;
 };
 
-private _muzzleList = GVAR(cachedThrowingWeapons);
-
-private _muzzleIdx = _muzzleList findIf {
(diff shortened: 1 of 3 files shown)


addons/main/functions/fnc_getCompatibleThrowMuzzle.sqf
```diff
@@ -0,0 +1,44 @@
...
+    private _compatible = GVAR(cachedThrowingWeaponsHash) get configName _x;
+    if (isNil "_compatible") then {
+        _compatible = getArray (_x >> "magazines");
+        GVAR(cachedThrowingWeaponsHash) set [configName _x, _compatible];
+    };
+    _magazine in _compatible
+};
+
+if (_muzzleIdx == -1) exitWith {
+    ""
```
diwako, line 41: shouldn't this also be cached? No muzzle found, cache it, so it does not need to iterate over everything again next time this is called?

jokoho48: yeah correct i totally fummbled the Caching Logic here... i should really not do 2 things at once and than try to split then into 2 PRs 😵‍💫

PR #368, opened 2024-01-10
WIP: Improve doCover logic

rekterakathom:
### When merged this pull request will: Make doCover activate for all danger sources and make it consider the direction of the danger.

Hopefully make firefights more dynamic by having the AI actually attempt to use cover by occasionally dashing behind soft- and hard cover, slightly biased towards the enemy so it'll "advance" towards the enemy.

- doCover might now activate for all sources of danger, including gunfire
- doCover now takes into account the direction of the danger to make units hide behind trees, rocks, etc
- doCover now has a slight bias for cover in front of the unit, so it might "advance" towards the enemy

I've declared this a draft, because I am not sure if this fits into lambs. It's pretty nice watching the AI now take cover, but I can see some issues, such as formations breaking, units wandering off, moving between two pieces of cover unnecessarily, etc. So, discussion is very much welcome.

### DIFF
--- a/addons/danger/functions/fnc_brainReact.sqf
+++ b/addons/danger/functions/fnc_brainReact.sqf
@@ -33,15 +33,15 @@ private _timeout = time + 1.4;
 // ACE3
 _unit setVariable ["ace_medical_ai_lastHit", CBA_missionTime];
 
-// cover move when explosion
+// Move to cover if it's relatively safe
 if (
-    getSuppression _unit < 0.6
+    getSuppression _unit < 0.4
+    && {RND(0.67)} // Add some randomness so they aren't taking cover all the time
     && {(speed _unit) isEqualTo 0}
-    && {_type in [DANGER_EXPLOSION, DANGER_FIRE]}
+    && {!(_unit call EFUNC(main,isIndoor))} // Don't try anything indoors
 ) exitWith {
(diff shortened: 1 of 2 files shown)


rautamiekka:
Fits perfectly, but the unwanted side effects are another story, especially if there just ain't a way to fix them without a huge Engine redesign. I suppose it could be a CBA setting.

rekterakathom:
Made an edit to clarify that I haven't personally seen any of the side effects I've described, but I can imagine them being possible. A lot of playtesting is definitely needed in many kinds of scenarios and maps. I do agree that this could maybe be in 2.6.1 as a CBA setting as an experimental feature so it can be widely tested.

nk3nny:
I think this feature works fine.  The only thing I would be somewhat concerned about is the change from Walk to Fast animations-- as these sometimes cause the AI to move through walls (as we see with doDodge).  With my limited testing I saw no issues.

I have generally avoided using _doMove_ where I can, because it means that player controlled groups will have some units "stuck" in cover (they won't automatically rejoin the player).  Here _moveTo_, while less reliable, does mean that the unit will try to return to formation.   This is for player controlled groups only.

nk3nny:
I like this feature.  I'm going to compile it into our internal modset and run it there. I'd also be happy to push it to dev for further testing.

PR #369, opened 2024-01-12
Improve anti-vehicle responses

### LINKED ISSUE #362: fnc_tactisHide anti-armour routines don't check launcher type
**Is your enhancement related to a problem?**

Potential in the current implementation for units to use non-anti-armor weaponry (or utility) against armor when it doesn't make any sense. **I haven't tested out if this actually happens in-game.**

**Solution you'd like:**

Anti-armor routines primarily look for units with anti-armor weapons. Currently it only looks for a launcher with no regard to the ammo type of the launcher. Launchers such as RHS's RShG-2 are intended for infantry units only, and should only be used against vehicles when no other choice is present. It is also possible that a launcher slot weapon is not a launcher at all, Prairie Fire has things like flare launchers.

So, my suggestion is as follows:
- Look for units with anti-armor launchers
- If not found, check for units with any launcher

**Alternatives you've considered:**

If this affects performance a lot, maybe just checking that it's actually a launcher is enough, to filter out any potentially stupid behaviour like units trying to shoot at tanks with flares.

**Additional context:**

I am specifically talking about L107 onwards of ``fnc_tacticsHide``

rekterakathom:
### When merged this pull request will: Add new function to find units with vehicle-capable launchers

Adds a new function that finds units that have vehicle-capable launchers (light AT, anti-air or heavy AT), and implements it into various anti-armor routines that previously only checked if units have secondary weapons. This is an important distinction, because various mods and CDLC's add secondary weapons that are not intended for AT usage. E.g, RHS's RshG-2 or the flare launchers in Prairie Fire.

The function can also account for submunitions, but this behaviour isn't used in this PR because I don't know if AI cares about the submunitions when deciding whether to engage. Disposable launchers from RHS for example are also completely ignored, but I don't consider this a huge flaw because in my testing AI refuses to use them anyway.

Also minor improvements in ``fnc_tacticsHide`` , such as making units target the vehicles themselves instead of crewmembers, and adding logic to ensure that AT units don't engage air vehicles and AA units don't engage ground vehicles.

Closes #362 

- Add new function ``lambs_main_fnc_getLauncherUnits``
- Implement ``getLauncherUnits`` into ``fnc_tacticsAssess``
- Implement ``getLauncherUnits`` into ``fnc_tacticsHide``
- Differentiate between AT and AA launchers in ``fnc_tacticsHide``
- Implement ``getLauncherUnits`` into ``fnc_taskRush``


addons/main/functions/fnc_getLauncherUnits.sqf
```diff
@@ -0,0 +1,73 @@
...
+    ["_offensiveAir", true, [true]],
+    ["_offensiveArmor", true, [true]]
+];
+
+private _suitableUnits = [];
+{
+    if ((secondaryWeapon _x) isEqualTo "") then {continue};
+    private _currentUnit = _x;
+
+    private _unitsMagazines = (magazines _x) + (secondaryWeaponMagazine _x);
```
rautamiekka, line 44:
```suggestion
    private _unitsMagazines = (magazines _currentUnit) + (secondaryWeaponMagazine _currentUnit);
```
Not an optimization but readability.

addons/main/functions/fnc_getLauncherUnits.sqf
```diff
@@ -0,0 +1,73 @@
...
+#define HEAVY_VEHICLE "512"
+
+params [
+    ["_group", grpNull, [grpNull]],
+    ["_checkSubmunition", false, [false]],
+    ["_offensiveVeh", true, [true]],
+    ["_offensiveAir", true, [true]],
+    ["_offensiveArmor", true, [true]]
+];
+
```
rautamiekka, line 38:
```suggestion
private _cfgMagazines = configFile >> "cfgMagazines";
private _cfgAmmo = configFile >> "cfgAmmo";

```

addons/main/functions/fnc_getLauncherUnits.sqf
```diff
@@ -0,0 +1,73 @@
...
+private _suitableUnits = [];
+{
+    if ((secondaryWeapon _x) isEqualTo "") then {continue};
+    private _currentUnit = _x;
+
+    private _unitsMagazines = (magazines _x) + (secondaryWeaponMagazine _x);
+    {
+        private _mainAmmo = getText (configFile >> "cfgMagazines" >> _x >> "ammo");
+        private _mainAmmoFlags = getText (configFile >> "cfgAmmo" >> _mainAmmo >> "aiAmmoUsageFlags");
```
rautamiekka, line 47:
```suggestion
        private _mainAmmo = getText (_cfgMagazines >> _x >> "ammo");
        private _mainAmmoFlags = getText (_cfgAmmo >> _mainAmmo >> "aiAmmoUsageFlags");
```

addons/main/functions/fnc_getLauncherUnits.sqf
```diff
@@ -0,0 +1,73 @@
...
+        if (
+            (_offensiveVeh && (LIGHT_VEHICLE in _mainAmmoFlags))
+            || {_offensiveAir && (AIR_VEHICLE in _mainAmmoFlags)}
+            || {_offensiveArmor && (HEAVY_VEHICLE in _mainAmmoFlags)}
+        ) exitWith {_suitableUnits pushBackUnique _currentUnit};
+
+        // Optionally go through submunitions. More info in header.
+        if !(_checkSubmunition) then {continue}; // Invert & continue to reduce indentation
+
+        private _submunition = getText (configFile >> "cfgAmmo" >> _mainAmmo >> "submunitionAmmo");
```
rautamiekka, line 58:
```suggestion
        private _submunition = getText (_cfgAmmo >> _mainAmmo >> "submunitionAmmo");
```

addons/main/functions/fnc_getLauncherUnits.sqf
```diff
@@ -0,0 +1,73 @@
...
+        if (_submunition isEqualTo "") then {continue};
+        private _submunitionFlags = getText (configFile >> "cfgAmmo" >> _submunition >> "aiAmmoUsageFlags");
```
rautamiekka, line 60:
```suggestion
        private _submunitionFlags = getText (_cfgAmmo >> _submunition >> "aiAmmoUsageFlags");
```

rautamiekka (changes requested):
Some config lookup optimizations, will save up more the bigger the squad.

addons/main/functions/fnc_getLauncherUnits.sqf
rekterakathom: Did you benchmark this? I benchmarked this, but the config lookups were so fast that making the paths into variables actually caused a slowdown. Something like ~0.220ms -> ~0.225ms. Tested on a vanilla NATO weapons squad, both without any changes and with the only launcher-equipped unit deleted, in both cases the version without the config path variables was faster.

A niche situation where the path variables might be faster is if every single squadmember has a secondary weapon that isn't suitable, but I think that's a very unconventional situation as squads by default rarely have more than 1 to 2 soldiers with a launcher, even in common faction mods.

rautamiekka: > Did you benchmark this?

Nope.

> making the paths into variables actually caused a slowdown.

Huh, interesting. Good to know.

nk3nny:
Interesting stuff. We'll be looking more closer at this.

addons/main/functions/fnc_getLauncherUnits.sqf
```diff
@@ -0,0 +1,73 @@
...
+
+private _suitableUnits = [];
+{
+    if ((secondaryWeapon _x) isEqualTo "") then {continue};
+    private _currentUnit = _x;
+
+    private _unitsMagazines = (magazines _currentUnit) + (secondaryWeaponMagazine _currentUnit);
+    {
+        private _mainAmmo = getText (configFile >> "cfgMagazines" >> _x >> "ammo");
+        private _mainAmmoFlags = getText (configFile >> "cfgAmmo" >> _mainAmmo >> "aiAmmoUsageFlags");
```
jokoho48, line 47: those values should probably be cached and additionally aiAmmoUsageFlags is not text but a number

addons/main/functions/fnc_getLauncherUnits.sqf
```diff
@@ -0,0 +1,73 @@
...
+            || {_offensiveAir && (AIR_VEHICLE in _mainAmmoFlags)}
+            || {_offensiveArmor && (HEAVY_VEHICLE in _mainAmmoFlags)}
+        ) exitWith {_suitableUnits pushBackUnique _currentUnit};
+
+        // Optionally go through submunitions. More info in header.
+        if !(_checkSubmunition) then {continue}; // Invert & continue to reduce indentation
+
+        private _submunition = getText (configFile >> "cfgAmmo" >> _mainAmmo >> "submunitionAmmo");
+        if (_submunition isEqualTo "") then {continue};
+        private _submunitionFlags = getText (configFile >> "cfgAmmo" >> _submunition >> "aiAmmoUsageFlags");
```
jokoho48, line 60: same as above aiAmmoUsageFlags is a number and also should be cached

jokoho48 (changes requested):
the aiAmmoUsageFlags and caching are from my pov here the biggest issue we should reuse the system that I implemented in #365. i also expect that that is the reason why you don't get RHS Disposable lauchers to work because aiAmmoUsageFlags is not a string but a number

addons/main/functions/fnc_getLauncherUnits.sqf
rekterakathom: aiAmmoUsageFlags is kind of both. For the purposes of this function, it is needed in text form.

``getText (configFile >> "CfgAmmo" >> "M_Vorona_HEAT" >> "aiAmmoUsageFlags") // Returns "128 + 512"``
``getNumber (configFile >> "CfgAmmo" >> "M_Vorona_HEAT" >> "aiAmmoUsageFlags") // Returns 640``

jokoho48: aiAmmoUsageFlags are bitflags and should also be used that way. this is also the reason why you don't get RHS Disposable and possibly other launchers because they may write use 640 for example

rekterakathom:
> the aiAmmoUsageFlags and caching are from my pov here the biggest issue we should reuse the system that I implemented in #365. i also expect that that is the reason why you don't get RHS Disposable lauchers to work because aiAmmoUsageFlags is not a string but a number

Not going to comment on the aiAmmoUsageFlags datatype, already mentioned in the conversation above.

The RHS disposable launchers might or might not work with this PR. What I meant in the original post was that RHS disposable launchers do not work with vanilla AI, they will not use them at all. Probably because they are technically empty until equipped, so the AI probably doesn't understand that it can be used.

I thought about making a cache for this, but config lookups here are so fast that it is a micro-optimization at best.

This takes ``~0.0030ms``, which goes down to ``~0.0027ms`` if the variables are removed
```
// Execute while holding a loaded launcher
private _mainAmmo = getText (configFile >> "cfgMagazines" >> (currentMagazine player) >> "ammo");
private _mainAmmoFlags = getText (configFile >> "cfgAmmo" >> _mainAmmo >> "aiAmmoUsageFlags");
_mainAmmoFlags
```

Simply calling something with parameters already takes up ``~0.0014ms``
```
["asdf", 0] call {params ["_string", "_number"]}
```

It is possible that the config lookups are only fast due to some config caching built into the engine that makes it fast in a benchmark like this where the same config gets accessed rapidly, but I still believe that the config lookups are not what is slowing this function down.

I didn't look into your caching system while the PR was still open, but now I did and I believe that it is currently bugged. 

Check the following input to your function:
```
["M_Vorona_HEAT", 2] call lambs_main_fnc_checkMagazineAiUsageFlags;
["M_Vorona_HEAT", 128] call lambs_main_fnc_checkMagazineAiUsageFlags
```
This will return ``false`` despite ``"M_Vorona_HEAT"`` having the 128 flag set. In addition to this, I don't believe that it would be much faster than the config lookups anyway, due to all the overhead that comes with calling it.

jokoho48:
This is not a topic of Performance but one of working correctly where we need always first to make sure that it does work correctly. and because aiAmmoUsageFlags is not a always a string or maybe a differently defined string we should always go for correctness.

and for the report that lambs_main_fnc_checkMagazineAiUsageFlags does not work. it needs to magazine not the ammo and `M_Vorona_HEAT` is the Ammo and not the Magazine is using `Vorona_HEAT` returns the expected Value

nk3nny:
I suspect you are talking around one-another. Nothing serious in the big picture-- though using the new function would put it more in line with the direction we are going now. 

Regardless. I'm having some trouble getting **lambs_main_fnc_getLauncherUnits** to correctly identify the TitanAA weapon carried by the humble "B_soldier_AA_F"  (NATO AA trooper).  It is quite late here, and I did test with ACE3. Perhaps I'm doing something wrong?

rekterakathom:
> This is not a topic of Performance but one of working correctly where we need always first to make sure that it does work correctly. and because aiAmmoUsageFlags is not a always a string or maybe a differently defined string we should always go for correctness.
> 
> and for the report that lambs_main_fnc_checkMagazineAiUsageFlags does not work. it needs to magazine not the ammo and `M_Vorona_HEAT` is the Ammo and not the Magazine is using `Vorona_HEAT` returns the expected Value

I seem to have some problem with my installation of the current lambs dev release because ``lambs_main_fnc_checkMagazineAiUsageFlags`` is nil for me, hence why I've declared it in this screenshot, but I have not altered it in any way other than removing the macros.

The first passed flag check is what gets cached, not the flags, hence the incorrect return in this screenshot.

> I suspect you are talking around one-another. Nothing serious in the big picture-- though using the new function would put it more in line with the direction we are going now.
> 
> Regardless. I'm having some trouble getting **lambs_main_fnc_getLauncherUnits** to correctly identify the TitanAA weapon carried by the humble "B_soldier_AA_F" (NATO AA trooper). It is quite late here, and I did test with ACE3. Perhaps I'm doing something wrong?

Tested, and indeed, ``lambs_main_fnc_getLauncherUnits`` fails. This is actually caused by the ammo usage flags being a number for ``M_Titan_AA``. ``lambs_main_fnc_checkMagazineAiUsageFlags`` would most likely fix this issue, once the cache is fixed. 

The pull request for that function has already been merged though so I can't make suggestions on it anymore. I would highly recommend that the function gets looked at thoroughly and actually tested, especially if its reason for existing is to be a reliable and correct way of getting the flags. In the future, consider practicing the same level of scrutiny for author pull requests as you do for non-author pull requests.

jokoho48:
Good catch https://github.com/nk3nny/LambsDanger/pull/382 should fix that. but the reason why you don't have the function in this PR is because you need to merge master 😄

rekterakathom:
> Good catch #382 should fix that. but the reason why you don't have the function in this PR is because you need to merge master 😄

I'll test that PR when I get done with school stuff for the day, but from a quick inspection it should work. I didn't test on this branch, just downloaded the dev version from steam and tested with that active, but steam had some trouble (as always) and apparently didn't update the mod so I was actually on 2.6.0 ...

And the change to ``lambs_main_fnc_checkMagazineAiUsageFlags`` on this branch coming soon™

rekterakathom:
``lambs_main_fnc_checkMagazineAiUsageFlags`` is now implemented, couldn't use it for the submunitions though. That would probably need yet another change to ``lambs_main_fnc_checkMagazineAiUsageFlags`` to make submunition checking possible through it, and I don't think that is required for 2.6.1 at least, it can be implemented later if the submunition checks are found to be of any use. So consider submunition stuff experimental and not-recommended for now.

This is currently mergeable with master branch but will be bugged until #382 is merged.

Verified that the Titan AA launcher is properly detected when this is used alongside #382.

addons/main/functions/fnc_getLauncherUnits.sqf
```diff
@@ -0,0 +1,72 @@
...
+    if ((secondaryWeapon _x) isEqualTo "") then {continue};
+    private _currentUnit = _x;
+
+    private _unitsMagazines = (magazines _currentUnit) + (secondaryWeaponMagazine _currentUnit);
+    {
+        if (
+            (_offensiveVeh && ([_x, LIGHT_VEHICLE] call lambs_main_fnc_checkMagazineAiUsageFlags))
+            || {_offensiveAir && ([_x, AIR_VEHICLE] call lambs_main_fnc_checkMagazineAiUsageFlags)}
+            || {_offensiveArmor && ([_x, HEAVY_VEHICLE] call lambs_main_fnc_checkMagazineAiUsageFlags)}
+        ) exitWith {_suitableUnits pushBackUnique _currentUnit};
```
jokoho48, line 50: Instead of calling the function 3x just add the values together and if the boolean is set and than only call checkMagazineAiUsageFlags once. additionally missing macros

rekterakathom: Fixed the macros. Hopefully I understood correctly what you meant with the 'add the values' bit.

jokoho48: yes you did but we should probably move the addition outside of the loops as from what I see they don't change

rekterakathom: Done

PR #374, opened 2024-01-28
Fix issues relating to constant dismounts

nk3nny:
This patches some of the issues pertaining to constant dismounting of units. 

It does this by rethinking how information is shared, and fixing a bug in doFleeing (which would not take into account the units current situation very well). Information is now only passed for units that are within "combatShareRange"  ~ 200 meters by default.  Units outside this distance, but within radio range, will unless busy, instead turn towards the direction of the unit calling for aid. 

Hiding is also tweaked to make that contact response more effective, which again translates into fewer repetitive executions of said command.

### DIFF
--- a/addons/main/functions/UnitAction/fnc_doFleeing.sqf
+++ b/addons/main/functions/UnitAction/fnc_doFleeing.sqf
@@ -14,7 +14,7 @@
  *
  * Public: No
 */
-#define SEARCH_FOR_HIDE 4
+#define SEARCH_FOR_HIDE 6
 #define SEARCH_FOR_BUILDING 8
 
 params ["_unit"];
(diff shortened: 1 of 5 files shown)


addons/main/functions/UnitAction/fnc_doFleeing.sqf
```diff
@@ -94,12 +96,13 @@ if (_onFootAndSeen) then {
 
     // find buildings to hide
     private _buildings = [_unit, SEARCH_FOR_BUILDING, true, true] call FUNC(findBuildings);
-    if ((_buildings isNotEqualTo []) && {_distance2D > random 5}) then {
+    _buildings append (_cover apply {getPos _x});
```
rekterakathom, line 99: ``getPos _x`` -> ``ASLtoAGL (getPosASL _x)``

Position format AGL is preferable for ``doMove``, and [getPos is a bad practice](https://community.bistudio.com/wiki/getPos)

nk3nny: We'll review that change later. (We like it).

diwako: We will be looking at that with the other pull request.

PR #380, opened 2024-02-18
Check callout validity

rekterakathom:

Check that callouts are valid when caching them.
Negligible performance degradation to caching logic (``~0.0012ms`` per check) but overall positive performance impact once cache is made.
This should remove all log errors relating to invalid files, e.g:
``Sound: Error: File: rhsafrf\addons\rhs_s_radio\radio\Male01\RU\Advance not found !!!``

- Check that callout sounds have file extensions when caching them

### DIFF
--- a/addons/main/functions/fnc_doCallout.sqf
+++ b/addons/main/functions/fnc_doCallout.sqf
@@ -103,7 +103,17 @@ if (_cachedSounds isEqualTo []) exitWith {
     };
 };
 
+// Not a valid file to play, extension must be included
 private _sound = selectRandom _cachedSounds;
+if !(".ogg" in _sound || ".wss" in _sound || ".wav" in _sound) exitWith {
+    if (GVAR(debug_functions)) then {
+        private _str = "WARNING: Callout file path %1 for callout %2 for speaker %3 is not valid!";
+        private _arr = [_str, _sound, _callout, _speaker];
+        _arr call FUNC(debugLog);
+        _arr call BIS_fnc_error;
+    };
+};
+
 playSound3D [_sound, _unit, isNull (objectParent _unit), eyePos _unit, 5, pitch _unit, _distance];
 [_unit, true] remoteExecCall ["setRandomLip", 0];
 [{
--- a/addons/main/functions/fnc_doCallout.sqf
+++ b/addons/main/functions/fnc_doCallout.sqf
@@ -77,13 +77,13 @@ if (isNil "_cachedSounds") then {
         private _sound = toLowerANSI _x; // playSound3D is case-insensitive
 
         // File extension must exist for playSound3D to work
-        if (_sound == "" || !(".ogg" in _sound || ".wss" in _sound || ".wav" in _sound)) then {
-            _sound = objNull;
-            _deleted = true;
-        } else {
+        if (_sound regexMatch ".+?\.(?:ogg|wss|wav|mp3)$/io") then {
             if (_sound select [0, 1] != "\") then {
                 _sound = (getArray (configFile >> "CfgVoice" >> _speaker >> "directories") select 0) + _sound;
             };
+        } else {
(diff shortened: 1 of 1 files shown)


addons/main/functions/fnc_doCallout.sqf
```diff
@@ -74,15 +74,18 @@ if (isNil "_cachedSounds") then {
     _cachedSounds = getArray (_protocolConfig >> _callout);
     private _deleted = false;
     {
-        private _sound = _x;
-        if (_sound == "") then {
+        private _sound = toLowerANSI _x; // playSound3D is case-insensitive
+
+        // File extension must exist for playSound3D to work
+        if (_sound == "" || !(".ogg" in _sound || ".wss" in _sound || ".wav" in _sound)) then {
```
rautamiekka, line 80: Considering `in` is rather slow on strings, RegEx might be faster (it's been faster before).

Alternatively, assuming there can't be anything after the extension for the sound to be considered legal, doing an end-of-string `select` may be even faster than a full-string RegEx.

rekterakathom: I think overengineering will result in a slowdown.

String to benchmark with (just smashed my keyboard):
``_sound = "a3479tghuaighreuyigvhbierlagheriap9tg745ra3heugjreivbugav7843fgrhbg864753g453hqb76453q9g54rgbh6q793g84gfcy43q.wav"``

Benchmarks:
``!(".ogg" in _sound || ".wss" in _sound || ".wav" in _sound);`` ~0.0013ms
``(_sound regexFind ["(.*?)\....", 0]) isEqualTo []`` ~0.0030ms

End-of-string select (not even doing the compare)
``_end = _sound select [((count _sound) - 4)]`` ~0.0013ms

An argument could be made for the RegEx, because even though it is slower, it will match any file extension. But I don't believe the engine even supports that many formats, mp3 maybe? But mp3 could just be added to the list of checks if it is supported.

rautamiekka: Try [`regexMatch`](https://community.bistudio.com/wiki/regexMatch) instead of [`regexFind`](https://community.bistudio.com/wiki/regexFind) if you don't need the match itself, and if you check just the extension it can be done by the `.+?\.(?:ogg|wss|wav)$/io` pattern (or `.+\.(?:ogg|wss|wav)$/io` if the former is somehow slower, or even more weirdly `.++\.(?:ogg|wss|wav)$/io`).

I admit I can't test it right now and [`select`](https://community.bistudio.com/wiki/select)'s docs make me suspicious, but it should support end-of-string/-array selections through negative indexes since 'Arma 3 v2.14':
```sqf
private _end = _sound select [-4]
```

nk3nny: As I recall, all callout sounds in vanilla are in .ogg

jokoho48: if we want to go after that, the fastest here is to check if a `.` is in the file path, though I don't think the speed is needed.
However, the question is if we maybe check if a file with the "supported" file endings exists so that we don't "filter" out still possible valid callouts and increase mod support.

rautamiekka: True.

rekterakathom: An alternative method would just be to run ``fileExists`` on it. That would guarantee that everything is valid, but it might be an order of magnitude slower. 

I'd appreciate it if somebody could run the benchmarks too for the current method, the ``regexMatch`` method and the ``fileExists`` method. I think ``regexMatch`` might be an improvement over the current one (even with a slight performance regression), as it's a much more compact statement and it's easier to add more filetypes (maybe mp3).

nk3nny:
Superseded by #397

PR #383, opened 2024-02-26
Fixes some instances of groups shooting into air

nk3nny:
Fixes Suppress, Flank and Assault group level actions -- where sometimes Units would shoot into air
Fixes units shooting UGL sometimes not partaking in assault.

### DIFF
--- a/addons/danger/functions/fnc_tacticsAssault.sqf
+++ b/addons/danger/functions/fnc_tacticsAssault.sqf
@@ -30,6 +30,9 @@ private _unit = leader _group;
 
 // find target
 _target = _target call CBA_fnc_getPos;
+if ((_target select 2) > 6) then {
+    _target set [2, 0.5];
+};
 
 // reset tactics
 [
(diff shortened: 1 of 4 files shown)


rekterakathom:
In other functions, instead of 'fixing' the height to 0.5, they just exit. Would that be a smarter thing to do here too? If a target is over 6 meters from the ground, what does it help to assume the height to be 0.5 instead and miss the target by at least 5.5 meters?

jokoho48:
> In other functions, instead of 'fixing' the height to 0.5, they just exit. Would that be a smarter thing to do here too? If a target is over 6 meters from the ground, what does it help to assume the height to be 0.5 instead and miss the target by at least 5.5 meters?

no, that would filter out the functions early as _target can be an "expected" position of a target as those are often Extrapolated based on last seen velocity

rautamiekka:
> it's better to sometimes not act even if a target is valid (extrapolation error) than to always act at the risk of unintended behaviour (shooting at least 5.5 meters below an actual target).

Inclined to agree that in terms of realism (and thus less immersion-breaking) it's better to ignore in this case than shoot God knows where; frozen from fear, a broken down gun, an impaired vision ... Any of just those would be a better flimsy excuse than shoot meters off the mark and claim a huge "Can't shoot straight" problem.

^ Still, if someone wanna play 'Arma', they can quickly disregard such immersion breakage if the mission still functions as intended, so I'm fine with either.

nk3nny:
Nice catch with **TacticsContact**. I'll update that too! As @jokoho48 notes the position passed to these functions are usually derived from _getHideFrom_.  This is the position the AI _thinks_ the enemy is occupying, but it unfortunately sometimes reads  far above ground. 

The benefit is that it gives, what I call in the design philosophy, the AI a degree of _fuzzy predictability_. The AI seems to be affected by the fog of war. The AI is operating not on perfect knowledge. And so on. You don't have to participate in many platoon level engagements with players, particularly PvP, to experience a lot of fire expended uselessly. ;) 

that said: I wouldn't be opposed with a function to replace our use of _getHideFrom_ with one that allows the option of giving the AI perfect getPosATL knoweldge(!).

PR #388, opened 2024-03-04
Improve vehicle maneuverability and defenses

nk3nny:
Adds Vehicle smoke usage when dropping off troops, or Jinking/dodging danger
Adds road checks to Assault, Flank and VehicleJink (to improve manoeuvrability)
Adds pseudo-fire team based bounds to infantry flanking together with Vehicles. 

Relies on #341


nk3nny:
Hemtt and linter problems are known. We'll discuss #341.

PR #389, opened 2024-03-04
Add features to fnc_tacticsReinforce.sqf

nk3nny:
Adds new features to tacticsReinforce

Adds groups set to reinforce will appropriate vehicles (of their side) to use for transport
Adds groups will take ownership of static weapons (and pack them for their use)
Adds groups will check dead bodies for useful weapon before rienforcing
Improves more reliable packing and deploying of static weapons
Improves more formations used
Fixes "HOLD", "GUARD", and "DISMISS" waypoints not being cancelled

Overall this is a considerable enhancement to how units with the reinforcement flag function.  They will much more reliably collect nearby resources before moving to reinforce, and add considerable dynamic potential to operations


nk3nny:
I wouldn't mind a discussion if there are more features that could be added here.

rautamiekka:
The recent updates already made the AI sound scarily realistic IIRC, and now even more. Gonna be interesting when we finally start playing again ... 😆

PR #391, opened 2024-03-04
Allows AI to target ACE3 incapacitated enemies in CQB

nk3nny:
**Potentially a controversial fix, but nonetheless grimly effective.** 

By allowing the AI to find and target ACE3 incapacitated enemies in CQB we FIX an inherent problem with the ACE3 system.  

Even though the incapacitated unit is rendered invisible, the FSM on an engine level still "knows" there is something there. But because the unit is invisible, it will merely start faffing about.  By allowing the AI to simply move there and kill the downed soldier, the problem resolves itself-- and the enemy becomes all the more deadly in CQB. 

Being incapacitated outside of CQB range is unchanged (and players and AI will still survive).

### DIFF
--- a/addons/danger/functions/fnc_brainEngage.sqf
+++ b/addons/danger/functions/fnc_brainEngage.sqf
@@ -50,6 +50,13 @@ if (
     && {(vehicle _target) isKindOf "CAManBase"}
     // && {_target call EFUNC(main,isAlive)}
 ) exitWith {
+
+    // set incapacitated ACE units visible
+    if (!GVAR(disableIncapacitatedKills)) then {
+        private _vis = _target getVariable ["ace_common_oldVisibility", 1];
+        if (_vis isEqualTo 0) then {_target setUnitTrait ["camouflageCoef", _vis];};
+    };
+
     _unit setVariable ["ace_medical_ai_lastFired", CBA_missionTime]; // ACE3
     [_unit, _target] call EFUNC(main,doAssault);
     _timeout + 1.4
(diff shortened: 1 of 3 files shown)


jokoho48:
This should probably be disabled by default, and/or we need to read out ace settings to check if an incapacitated unit can die because else we are back at the point of Mag Dumping into incapacitated players/units that we had before

rekterakathom:
If this gets added, I definitely agree with Joko that this should be disabled by default, I personally feel like AI magdumping unconscious people is very irritating, and deducts from gameplay. If an unconscious player gets magdumped, then any medical gameplay will just go out the window because nobody will bother healing anyone with 30 avulsions in their torso.

A3 2.18 will add a new command, [ignoreTarget](https://community.bistudio.com/wiki/ignoreTarget) that could be a more elegant solution, perhaps give it a go on the dev branch?

addons/danger/functions/fnc_brainEngage.sqf
```diff
@@ -50,6 +50,13 @@ if (
     && {(vehicle _target) isKindOf "CAManBase"}
     // && {_target call EFUNC(main,isAlive)}
 ) exitWith {
+
+    // set incapacitated ACE units visible
+    if (!GVAR(disableIncapacitatedKills)) then {
+        private _vis = _target getVariable ["ace_common_oldVisibility", 1];
```
diwako, line 56: Are you sure `_target` is local here? `ace_common_oldVisibility` is only available where `_target` is local, `setUnitTrait` also requires the `_target` to be local and will globally propagate the camouflage change.

PR #395, opened 2024-05-14
Improve GetLauncherUnits

jokoho48:

1. *Depends on #369*
2. *Simplify by using flags directly instead of bool flags*
3. *Add AI_AMMO_USAGE_FLAG_X Macros for easier use*


addons/main/functions/fnc_getLauncherUnits.sqf
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * Example:
+ * [group bob] call lambs_main_fnc_getLauncherUnits;
+ *
+ * Public: Yes
+*/
+
+params [
+    ["_group", [], [grpNull, []]],
+    ["_flags", 896, [0]], // 896 == AI_AMMO_USAGE_FLAG_VEHICLE + AI_AMMO_USAGE_FLAG_AIR + AI_AMMO_USAGE_FLAG_ARMOUR
```
diwako, line 30: why not just do `AI_AMMO_USAGE_FLAG_VEHICLE + AI_AMMO_USAGE_FLAG_AIR + AI_AMMO_USAGE_FLAG_ARMOUR` instead of 896?

addons/main/functions/fnc_getLauncherUnits.sqf
```diff
@@ -0,0 +1,63 @@
...
+
+if (_group isEqualType grpNull) then {
+    _group = units _group;
+};
+private _suitableUnits = [];
+{
+    if ((secondaryWeapon _x) isEqualTo "") then {continue};
+    private _currentUnit = _x;
+
+    private _unitsMagazines = (magazines _currentUnit) + (secondaryWeaponMagazine _currentUnit);
```
diwako, line 42: best include `handgunMagazine` as well, RHs for example has a grenade launcher for the pistol slot.

jokoho48: it checks in L39 explicitly for secondaryWeapon so we can ignore that I would say

diwako: actually hold up, secondary weapon in arma terms is always the launcher slot.

jokoho48: yes that is correct we don't check for hand guns in this function

rekterakathom: > best include `handgunMagazine` as well, RHs for example has a grenade launcher for the pistol slot.

In the context that this function is used, I don't think grenade launchers should be detected. This function is used to detect anti-vehicle launchers to identify which units should be given orders to destroy vehicles. 40mm HEDP arguably could fit into this category, but I believe it's best for code simplicity to not bother with that. I think the AI will fire 40mm grenades into soft vehicle targets even without being explicitly told to do so.

PR #396, opened 2024-05-20
Bump REQUIRED_VERSION

rekterakathom:

Set the correct required version. I believe the latest command on master is ``getOrDefaultCall`` which was introduced in A3 version 2.12

Title

The whole version check might be [disabled(?)](https://community.bistudio.com/wiki/CfgPatches) but it's good practice to keep it up-to-date I suppose.


jokoho48:
Maybe we should just push it to the latest(2.16) instead of the "minimum Required" version

rekterakathom:
> Maybe we should push it to the latest(what should be 2.16) instead of the "minimum Required" version

Not a bad idea either. Maybe edit this PR to match the latest version when 2.6.1 gets released, so 2.16 or 2.18, and then merge.

PR #397, opened 2024-05-21
Use file exists for do callout

jokoho48:

1. *Possible Replacement for #380*

It is generally slower than #380, but it does not matter because it gets cached for the benefit of fewer false discards.

### DIFF
--- a/addons/main/functions/fnc_doCallout.sqf
+++ b/addons/main/functions/fnc_doCallout.sqf
@@ -79,8 +79,14 @@ if (isNil "_cachedSounds") then {
             _cachedSounds set [_forEachIndex, objNull];
             continue;
         };
+        
+        private _hasFileEnding = _sound regexMatch ".+?\.(?:ogg|wss|wav|mp3)$/io";
 
-        if !(fileExists _sound) then {
+        if (_sound select [0, 1] != "\") then {
+            _sound = (getArray (configFile >> "CfgVoice" >> _speaker >> "directories") select 0) + _sound;
+        };
+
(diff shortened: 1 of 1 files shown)


addons/main/functions/fnc_doCallout.sqf
```diff
@@ -74,15 +73,33 @@ if (isNil "_cachedSounds") then {
...
+            continue;
         };
+        
+        private _hasFileEnding = _sound regexMatch ".+?\.(?:ogg|wss|wav|mp3)$/io";
+
+        if (_sound select [0, 1] != "\") then {
+            _sound = (getArray (configFile >> "CfgVoice" >> _speaker >> "directories") select 0) + _sound;
+        };
+
+        if (!fileExists _sound && !_hasFileEnding) then {
```
rekterakathom, line 89: Switch condition order and add lazy eval for an easy optimization, ``fileExists`` is a lot slower than checking the variable

``if (!_hasFileEnding && {!fileExists _sound}) ..``

jokoho48: good catch

rekterakathom:
Looks good. I think this one manages to find RHS callouts by adding the file extension, which is something that #380 can't do.

I added the ``.mp3`` extension to #380 "just to be sure" since it didn't hurt performance, but here it might have a big impact on the worst-case performance (when all extensions are tested). Perhaps it's worth testing whether or not ``.mp3`` is a valid format accepted by the engine at all, [the biki says it is not](https://community.bistudio.com/wiki/WSS_File_Format).

PR #398, opened 2024-05-26
Add Deprecated to Vanilla Waypoints

jokoho48:

1. *Title*

### DIFF
--- a/addons/wp/stringtable.xml
+++ b/addons/wp/stringtable.xml
@@ -1221,5 +1221,59 @@
             <Italian>Registri artiglieria per uso dinamico</Italian>
             <Chinesesimp>动态使用方面的火炮指示</Chinesesimp>
         </Key>
+        <Key ID="STR_Lambs_WP_Waypoint_Deprecated_TaskAssault_displayName">
+            <English>!!!DEPRECATED, USE MODULES!!! - TASK ASSAULT</English>
+        </Key>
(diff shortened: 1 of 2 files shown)


diwako:
Needs info in which version they are removed.

jokoho48:
> Needs info in which version they are removed.

We can't remove them without breaking Missions, so I think we may just mark them as "deprecated" and leave them.

addons/wp/stringtable.xml
```diff
@@ -1221,5 +1221,59 @@
             <Italian>Registri artiglieria per uso dinamico</Italian>
             <Chinesesimp>动态使用方面的火炮指示</Chinesesimp>
         </Key>
+        <Key ID="STR_Lambs_WP_Waypoint_Deprecated_TaskAssault_displayName">
+            <English>TASK ASSAULT</English>
+        </Key>
+        <Key ID="STR_Lambs_WP_Waypoint_Deprecated_TaskAssault_Tooltip">
+            <English>Rushes to target position with no regard for safety\nDEPRECATED, USE MODULES FOR MORE FLEXIBILITY FOR MORE FLEXIBILITY</English>
```
johnb432, line 1228: `USE MODULES FOR MORE FLEXIBILITY FOR MORE FLEXIBILITY` - looks like a duplicate snuck in there.

jokoho48: i blame @nk3nny

PR #399, opened 2024-05-28
Revert change in fnc_doCallout

rekterakathom:
Reverts a part of #355 because of recent discovery that files included in addons must **not** have a leading slash ([source](https://discord.com/channels/105462288051380224/105465701543723008/1245083642557104338)). Haven't verified this myself but this has basically no performance impact due to caching so might as well revert it just to be safe.

This will conflict with #380 and #397, I'll fix it when the time comes.

Removes leading slash from callout file paths

Removes leading slash from callout file paths


jokoho48:
when Merge Conflicts are resolved this can get merged

rekterakathom:
Merge conflict should be resolved now

PR #402, opened 2024-07-07
Improve ExplosionEH animations with CBA Events

nk3nny:
Adds removeExec to animations on explosions to ensure they actually trigger.


rekterakathom:
Maybe worth testing the ``switchMove`` alt syntax on dev branch to see if that syntax can force instant transitions on remote clients without remote execution to avoid unnecessary network traffic.

nk3nny:
Good point.

nk3nny:
Change to CBA eventhandlers

PR #407, opened 2024-07-14
Fix BrainForced used PRONE instead of DOWN

### LINKED ISSUE #406: Unknown enum value: "PRONE"
**Mods (complete and add to the following information):**

- **Arma 3:** `2.14 stable`
- **CBA:** `3.17.1`
- **LAMBS_DANGER:** `2.6.0`

**Description:**

RPT and on-screen error caused by an unknown enum value

**Steps to reproduce:**

- Unknown. Only received reports from others about it. Provable issue via bi wiki documentation.

**Expected behavior:**

Error does not occur

**Where did the issue occur?**

Dedicated Multiplayer

**Additional context:**

https://github.com/nk3nny/LambsDanger/blob/9670c56ef4ed48716387974abf278fe3faf89a83/addons/danger/functions/fnc_brainForced.sqf#L55

[BI Wiki](https://community.bistudio.com/wiki/setUnitPosWeak)

**Log Files:**
RPT error:
```
9:50:16 Error in expression <ll lambs_main_fnc_doAssaultSpeed;
_unit setUnitPosWeak (["MIDDLE", "PRONE"] sele>
 9:50:16   Error position: <setUnitPosWeak (["MIDDLE", "PRONE"] sele>
 9:50:16   Error Foreign error: Unknown enum value: "PRONE"
 9:50:16 File z\lambs\addons\danger\functions\fnc_brainForced.sqf..., line 55
 9:50:16 [ACRE] (sys_io) INFO: Pipe error: Read CreateFileA WinErrCode: 2
```

(I think the fix is replacing the enum with `"DOWN"` but I don't have time to read the full code context)

jokoho48:

1. *Fixes #406*

### DIFF
--- a/addons/danger/functions/fnc_brainForced.sqf
+++ b/addons/danger/functions/fnc_brainForced.sqf
@@ -52,7 +52,7 @@ if ((currentCommand _unit) isEqualTo "ATTACK") then {
         _unit doWatch objNull;
     };
     [_unit, _attackTarget] call EFUNC(main,doAssaultSpeed);
-    _unit setUnitPosWeak (["MIDDLE", "PRONE"] select (getSuppression _unit > 0.9));
+    _unit setUnitPosWeak (["MIDDLE", "DOWN"] select (getSuppression _unit > 0.9));
     _unit setVariable [QEGVAR(main,currentTask), "Attacking", EGVAR(main,debug_functions)];
 };
 


diwako (approved):

rekterakathom:
I don't know if this pull request is the right place to discuss this, but oh well:

Should this behavior just be removed instead of fixed? Fixing this error will change the behavior of units even though it has been this way for 3 years now, and I don't see anything wrong with the way it has been.

A lot of people (including myself) don't like that the AI goes prone when it shouldn't and I feel like "fixing" this would just increase the amount of weird crawling around that the AI already does.

Havoc-1:
The AI having a sense of self-preservation actually contributes to the gameplay experience positively and I would not mind having this merged with the release branch. 

_A lot of people (including myself)_ like the AI going prone when they take fire (just as players do). In the best interest of both parties it might be worth considering to make it togglable via CBA settings to cater to people of different playstyles and preferences.

diwako:
Today we will be discussing this PR internally, mostly on the basis of changing the current AI behavior or fixing a long standing bug that was just never caught.

nk3nny (approved):

nk3nny:
@rekterakathom 
The change fixes a bug.  It doesn't affect general behaviour, only in a narrow set of circumstances when the AI is in the "attack" fsm (actively hunting a target) and under significant suppression. 

This will increase survivability-- but should not affect the AI when clearing rooms or the like.

PR #410, opened 2024-08-25
change toLower to toLowerANSI

jokoho48:

1. *Title*

### DIFF
--- a/addons/main/XEH_postInit.sqf
+++ b/addons/main/XEH_postInit.sqf
@@ -11,7 +11,7 @@ GVAR(CalloutCacheNamespace) = call CBA_fnc_createNamespace;
         if !(".p3d" in _string) then {
             _string = _string + ".p3d";
         };
-        GVAR(blockSuppressionModelCache) setVariable [toLowerANSI _string, true];
+        GVAR(blockSuppressionModelCache) setVariable [toLower _string, true];
     };
 } forEach ("private _name = configName _x; _name isKindof 'building' || {_name isKindOf 'Rocks_base_F'}" configClasses (configFile >> "CfgVehicles"));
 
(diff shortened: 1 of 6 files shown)


diwako:
Huh? Why?

jokoho48:
https://community.bistudio.com/wiki/toLowerANSI
> Converts the supplied string to all lower case characters fast (could be 3× faster than [toLower](https://community.bistudio.com/wiki/toLower)).

addons/main/XEH_postInit.sqf
```diff
@@ -11,7 +11,7 @@ GVAR(CalloutCacheNamespace) = call CBA_fnc_createNamespace;
         if !(".p3d" in _string) then {
             _string = _string + ".p3d";
         };
-        GVAR(blockSuppressionModelCache) setVariable [toLower _string, true];
+        GVAR(blockSuppressionModelCache) setVariable [toLowerANSI _string, true];
```
diwako, line 14: Here we probably should not do that, in case there is someone using none standard ANSI characters for the file path.

jokoho48: As far as I remember, setVariable only supports ASCII characters.

diwako:
```sqf
private _yeet = false call CBA_fnc_createNamespace; 
private _string = "\правда\model.p3d"; 
if ((_string select [0,1]) isEqualTo "\") then { 
 _string = [_string, 1] call CBA_fnc_substr; 
}; 
if !(".p3d" in _string) then { 
 _string = _string + ".p3d"; 
}; 
_yeet setVariable [toLowerANSI _string, true]; 
allVariables _yeet
```

returns `["������������\model.p3d"]`

diwako: that means obfuscated pbos will not work well here.

PR #418, opened 2024-10-18
Eventhandlers - Add Delay to AI Explosion Reaction

lambdatiger:

1. Add a delay to explosionEH dependent on:
   - Distance from AI
   - Auditory response times
   - AI Skill

### DIFF
--- a/addons/eventhandlers/functions/fnc_delayExplosionEH.sqf
+++ b/addons/eventhandlers/functions/fnc_delayExplosionEH.sqf
@@ -13,7 +13,7 @@
  * None
  *
  * Example:
- * [BIS EXPLOSION EH ARGS] call lambs_eventhandlers_fnc_explosionEH;
+ * [BIS EXPLOSION EH ARGS] call lambs_eventhandlers_fnc_delayExplosionEH;
  *
  * Public: No
 */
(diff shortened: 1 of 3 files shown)


nk3nny:
I love this. The effect is slight, _very slight_, but noticable when you know what to look for.  I wonder if the logic of sound/distance can be used elsewhere too.

addons/eventhandlers/functions/fnc_delayExplosionEH.sqf
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Public: No
+*/
+params [
+    ["_unit", objNull],
+    "",
+    ["_explosionSource", objNull]
+];
+
+if (!alive _unit ||
```
jokoho48, line 26: we should move/do more of the checks we do in the [here](https://github.com/nk3nny/LambsDanger/blob/master/addons/eventhandlers/functions/fnc_explosionEH.sqf#L22) before we run the waitAndExecute

jokoho48 (changes requested):

addons/eventhandlers/functions/fnc_delayExplosionEH.sqf
lambdatiger: I agree, but I think a few of these could change in time:
- `!GVAR(ExplosionEventHandlerEnabled)` - should have been there
- `!alive` - covers `isNull` - probably could be replaced with `{!(_unit call EFUNC(main,isAlive))`
- `{!isNul objectParent _unit}` - could be added, but units may be getting in/out and the transition happens in time
- `{(stance _unit) isEqualTo "PRONE"}` - a thing that could be changing in that ~0.15-0.3 ms
- `{_unit getVariable [QGVAR(explosionReactionTime), 0] > time}` -  delay could be computed and then checked against this statement

I'll make the changes I think I made clear I think are good ideas, but for checking if a unit is in a vehicle or is prone I could go either way on.

PR #421, opened 2024-11-10
Improvements to brainEngage

nk3nny:
Improve **brainEngage** by taking advantage of _getUnitState_
Improve **brainEngage** reimplements behaviour not triggering when **stealthed***
Fixes _doAssault_ where a unit would sometimes get stuck trying to check a position that the group believed the enemy is-- where it clearly could not be. (such as in the sky)**

*Not triggering _brainEngage_ when a unit is sneaking means that the AI become more predictable for player led groups, and unit set to ambush actually remain-ambushing.  Those set to sneak will not get mixed commands from their dangerAI

**It does this by simply forgetting the enemy, which is somewhat controversial but works very well in the context of group memory (i.e., if it is near a building, it will still have likely places to check). 

Makes me think we really should make our own _getHideFrom_ function.

### DIFF
--- a/addons/danger/functions/fnc_brainEngage.sqf
+++ b/addons/danger/functions/fnc_brainEngage.sqf
@@ -4,7 +4,7 @@
  * handles responses while engaging
  *
  * Arguments:
- * 0: unit doing the avaluation <OBJECT>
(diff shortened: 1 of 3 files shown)


nk3nny:
Yeah. Linter will yell

PR #422, opened 2024-11-10
Improve Dodging and reactions

nk3nny:
Take advantage of new _getUnitState_ command

Fix some cases of units warping through cover
Improve Dodge and Cover conditions. Units are more likely to drop prone and appear suppressed if under heavy fire.


nk3nny:
<3 linter.

PR #428, opened 2024-11-23
Adds teleport option to taskPatrol

nk3nny:
Extends Modules (Zeus and Eden) with a teleport option in taskPatrol.  When enabled, it functions much like Garrisons-- placing the group at at a randomly selected position with the patrol area.


addons/wp/functions/Modules/fnc_modulePatrol.sqf
```diff
@@ -34,13 +34,14 @@ switch (_mode) do {
                         [LSTRING(Module_TaskPatrol_Range_DisplayName), "SLIDER", LSTRING(Module_TaskPatrol_Range_ToolTip), [20, 2000], [1, 0.5], TASK_PATROL_SIZE, 1],
                         [LSTRING(Module_TaskPatrol_Waypoints_DisplayName), "SLIDER", LSTRING(Module_TaskPatrol_Waypoints_ToolTip), [2, 15], [2, 1], TASK_PATROL_WAYPOINTCOUNT, 0],
                         [LSTRING(Module_TaskPatrol_MoveWaypoints_DisplayName), "BOOLEAN", LSTRING(Module_TaskPatrol_MoveWaypoints_ToolTip), TASK_PATROL_MOVEWAYPOINTS],
-                        [LSTRING(Module_Task_EnableReinforcement_DisplayName), "BOOLEAN", LSTRING(Module_Task_EnableReinforcement_ToolTip), TASK_PATROL_ENABLEREINFORCEMENT]
+                        [LSTRING(Module_Task_EnableReinforcement_DisplayName), "BOOLEAN", LSTRING(Module_Task_EnableReinforcement_ToolTip), TASK_PATROL_ENABLEREINFORCEMENT],
+                        [LSTRING(Module_TaskGarrison_Teleport_DisplayName), "BOOLEAN", LSTRING(Module_TaskGarrison_Teleport_Tooltip), TASK_PATROL_TELEPORT]
```
jokoho48, line 38: When we use it in multiple places, we should remove the Garrison from the name

addons/wp/functions/fnc_taskPatrol.sqf
```diff
@@ -12,6 +12,7 @@
  * 4: Area the AI Camps in, default [] <ARRAY>
  * 5: Dynamic patrol pattern, default false <BOOL>
  * 6: enable dynamic reinforcement <BOOL>
+ * 7: Teleport group <BOOL>
```
rautamiekka, line 15: This needs far better wording. The [comment](https://github.com/nk3nny/LambsDanger/pull/428/files#diff-a7c7c916574286f9ebceb75bd0d1ec5b509b333c885b6dc9b10c0d8b5ddc5840R123) for the associated code says it teleports the group to a random waypoint position, which itself is confusing cuz it means a random waypoint from the ones the group currently has.

rautamiekka (changes requested):

addons/wp/functions/Modules/fnc_modulePatrol.sqf
nk3nny: Good call

addons/wp/functions/fnc_taskPatrol.sqf
nk3nny: Good point!

PR #429, opened 2024-11-24
Add Defense Module

nk3nny:
Adds a new **scripted Defense module** with Eden, Zeus and Zen integration. 

The module adds a way to create a defensive spot and have the a group Defend and Hold that position exclusively.  The module can be configured for range, teleporting into place,  random area patrols, and preferred cover positions. 

Using module settings to decide if the group should prefer buildings, walled fortifications, vegetations, or any combination to use as cover positions.  Once engaged, the group will remain within the area moving from cover to cover according to proximity to enemy. 

The group will also frequently share information and attempt to call for Artillery fire. Making it an ideal function for those kinds of uses.  Using the "stealth" or "ambush" flag setup defensive ambushes. Also excellent to model artillery spotters that will hold fire, but actively call for artillery strikes against spotted enemies.


nk3nny:
**An Ode to Stringtables**

Oh stringtables, thou art my constant muse,
In thee resides the beauty of pure text.
Thy structure, neat, doth chaos oft diffuse,
A scaffold firm where logic is annexed.

Thy keys, like faithful messengers, do bear
The weight of meaning, clear and unconfused.
Each string, a jewel, thy table doth prepare,
A language tapestry, so finely used.

When scripts grow wild and languages divide,
Thy order stands, a bridge across the seas.
With every lookup, thou dost true abide,
Uniting code and word with fluent ease.

Oh, sacred stringtables, thy worth I sing,
For in thy rows, a coder’s heart takes wing.

PR #434, opened 2025-01-18
Improve Hiding feature

nk3nny:
This pull request improves hiding by streamlining the decision process and removing some of the forced movement-- which in turn makes the AI more survivable. 

Units will prioritise terrain cover more than before-- which in turn makes the AI faster when trying for cover. Likewise each individual unit will be in charge of its own seeking cover, which makes the hiding again faster and more reliable. 

Added a new Anti-tank/air contact state
Improved anti-air/tank tactics in tacticsHide
Improved  streamlined doGroupHide.sqf

### DIFF
--- a/addons/main/functions/UnitAction/fnc_doHide.sqf
+++ b/addons/main/functions/UnitAction/fnc_doHide.sqf
@@ -13,7 +13,7 @@
  * boolean
  *
  * Example:
- * [bob, angryJoe, 30] call lambs_main_fnc_doHide;
+ * [bob, angryJoe] call lambs_main_fnc_doHide;
  *
  * Public: No
 */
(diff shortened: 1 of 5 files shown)


addons/main/functions/UnitAction/fnc_doHide.sqf
```diff
@@ -27,7 +27,8 @@ if (
 ) exitWith {false};
 
 // do nothing when already inside
-if (RND(GVAR(indoorMove)) && {_unit call FUNC(isIndoor)}) exitWith {
+if (RND(GVAR(indoorMove)) && { lineIntersects [ eyePos _unit, ( eyePos _unit ) vectorAdd [0, 0, 10] ] } ) exitWith {
```
rautamiekka, line 30: Knowing even the server part doesn't run extremely well (especially if not a dedicated, and even more especially if on the same machine as the client), saving the `eyePos` probly would make it faster and probly avoid an instance where both statements differ:
```suggestion
if (RND(GVAR(indoorMove)) && { private _eyePos = eyePos _unit; lineIntersects [ _eyePos, _eyePos vectorAdd [0, 0, 10] ] } ) exitWith {
```

rautamiekka (changes requested):

PR #438, opened 2025-01-23
Update fnc_taskDefend.sqf

nk3nny:
Improved prevention of unit stacking
Fixed using small radius zones or limiting to buildings only
Fixed units teleporting on top of buildings


addons/wp/functions/fnc_taskDefend.sqf
```diff
@@ -261,21 +269,23 @@ private _handle = [
                 _defensivePos = _defensivePos apply {_x select 1};
 
                 // check available cover
-                if (count _units > count _defensivePos) then {_units = _units resize (count _defensivePos)};
+                if (count _units > count _defensivePos) then {_units resize (count _defensivePos)};
```
rautamiekka, line 272: That was a funny mistake, ngl.

PR #441, opened 2025-02-17
Add group settings for Attack, Reinforce and Dynamic Reinforcement

nk3nny:
Adds universal group settings for Attack, Flight and Dynamic reinforcement

Disabling '_enableattack_' on groups has the benefit of giving us more control of group actions (less running away doing random shit).  The cost is that the AI will conduct fewer unit level flanking attacks.

### DIFF
--- a/addons/danger/XEH_preInitClient.sqf
+++ b/addons/danger/XEH_preInitClient.sqf
@@ -10,7 +10,7 @@ Here four buttons are added:
 These functions and buttons could conceivably be moved to their own module.
 They might also benefit from added interface and more nuanced functions handling.
 That said: within the current scope of the AI mod. They can make a comfortable
-home here. Until revisited by more capable personnell.
+home here. Until revisited by more capable personell.
 - nkenny
 */
 
@@ -103,7 +103,7 @@ private _fnc_assault_AI = {
     } else {
         _cursorPos = getPos cursorObject;
     };
-    private _buildings = [_cursorPos, 50, true, false] call EFUNC(main,findBuildings);
+    private _buildings = [_cursorPos, 12, true, false] call EFUNC(main,findBuildings);
     (group player) setVariable [QEGVAR(main,groupMemory), _buildings];
     {
         private _enemy = _x findNearestEnemy _cursorPos;
--- a/addons/danger/settings.inc.sqf
+++ b/addons/danger/settings.inc.sqf
@@ -108,6 +108,70 @@ _curCat = LSTRING(Settings_GeneralCat);
     1
 ] call CBA_fnc_addSetting;
 
+addMissionEventHandler ["GroupCreated", {
+    params ["_group"];
+    [
+        {
+            if (count (units _this) > 0) then {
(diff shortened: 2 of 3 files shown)


addons/danger/settings.inc.sqf
```diff
@@ -108,6 +108,70 @@ _curCat = LSTRING(Settings_GeneralCat);
     1
 ] call CBA_fnc_addSetting;
 
+addMissionEventHandler ["GroupCreated", {
```
jokoho48, line 111: This should probably be in the EventHandlers pbo

addons/danger/settings.inc.sqf
```diff
@@ -108,6 +108,70 @@ _curCat = LSTRING(Settings_GeneralCat);
...
+    params ["_group"];
+    [
+        {
+            if (count (units _this) > 0) then {
+                if (GVAR(autoAddDynamicReinforcement)) then {_this setVariable [QGVAR(enableGroupReinforce), true, true];};
```
jokoho48, line 116: i think we maybe should split this up into a more logical system. why should we, for example, wait when all 3 settings are disabled?

addons/danger/XEH_preInitClient.sqf
```diff
@@ -10,7 +10,7 @@ Here four buttons are added:
 These functions and buttons could conceivably be moved to their own module.
 They might also benefit from added interface and more nuanced functions handling.
 That said: within the current scope of the AI mod. They can make a comfortable
-home here. Until revisited by more capable personnell.
+home here. Until revisited by more capable personell.
```
rautamiekka, line 13: As far as I can find, the former `personnell` was closer, should just drop the extra trailing `l`. Still, I'd go with `people` instead, since seemingly `personnel` is people working in a company:
```suggestion
home here. Until revisited by more capable people.
```

addons/danger/settings.inc.sqf
rautamiekka: It's a 3-second wait before executing ([`waitAndExecute`](https://github.com/CBATeam/CBA_A3/blob/master/addons/common/fnc_waitAndExecute.sqf)) instead of waiting until all 3 are enabled ([`waitUntilAndExecute`](https://github.com/CBATeam/CBA_A3/blob/master/addons/common/fnc_waitUntilAndExecute.sqf)), but that's still a good point.

rautamiekka (changes requested):

addons/danger/settings.inc.sqf
```diff
@@ -108,6 +108,70 @@ _curCat = LSTRING(Settings_GeneralCat);
...
+        if (_value) exitWith {};
+        if (!_value && time > 0) then {
+            {_x enableAttack false;} forEach (allGroups select {local (leader _x)});
+        };
+    }
+] call CBA_fnc_addSetting;
+
+// auto disable allowFleeing on groups
+[
+    QGVAR(disableFleeingOverride),
```
diwako, line 144: This and the next setting need a XEH initPost eventhandler instead. People spawn in units all the time, which will not get caught by this.

PR #445, opened 2025-09-20
Remove contact sub-states

nk3nny:
Removes the contact sub-states. After a serious round of testing, I have ended up removing the contact-phase substates (i.e., contact-assaulting, contact-suppressing, contact-hiding, etc.).

These did little but muddle up the AIs reactions, and remove focus from what was supposed to be the real group level tactics-- namely the group manoeuvres. This has the massive benefit of not interrupting the core _LAMBS Danger.fsm_ functions in their most important moments. 

Additionally there are tweaks to hiding and assessment states to take advantage of the streamlined performance-- while improving on the functionality given through the group level tactics. 

**Highlights:** 
- Simplifies AI responses in a new contact
- Streamlines leadership checks
- Tweaks assessment and hiding 
- Renames **fnc_tacticsContact** to **fnc_contact** to distinguish the split.

### DIFF
--- a/tools/sqf_linter_LogChecker.py
+++ b/tools/sqf_linter_LogChecker.py
@@ -1,7 +1,7 @@
 import sys
 import os
 
-defaultFalsePositives = 33
+defaultFalsePositives = 31
 def main():
     f = open("sqf.log", "r")
     log = f.readlines()


addons/danger/functions/fnc_tacticsAssess.sqf
```diff
@@ -146,16 +146,16 @@ if !(_enemies isEqualTo [] || {_unitCount < random 4}) then {
...
-    private _farHighertarget = _enemies findIf {
+    private _farHigherTarget = _enemies findIf {
         !_speedMode
         && {_unit distance2D _x > RANGE_LONG}
         && {_unit knowsAbout _x < 2}
         && {((getPosASL _x) select 2) > ((_eyePos select 2) + 15)}
         && {!(terrainIntersectASL [_eyePos vectorAdd [0, 0, 5], eyePos _x])}
     };
-    if (_farHighertarget != -1) exitWith {
+    if (_farHigherTarget != -1) exitWith {
```
rautamiekka, line 156:
```suggestion
    if (_farHigherTarget isNotEqualTo -1) exitWith {
```

rautamiekka (changes requested):
Seems fine to me, but I wasn't there, so I can only look for glaring issues.

Though, personally would truncate the state values to the absolute bare minimum, especially if there's a chance it's gonna be broadcast.
- A single letter with the 26 values per casing (`A`-`Z`, `a`-`z`), or, hell, even a single integer with 10 values (`0`-`9`), would allow for so many different states that even LAMBS couldn't use them all, especially now that the system is being simplified. Whether broadcasting a string of an uppercase letter (https://www.sciencebuddies.org/science-fair-projects/references/ascii-table makes it seem only the decimal form of the lowercase letters might take more space, but I didn't search more and thus could be wrong) or an integer as is or a string version of an integer is more efficient is beyond me.

jokoho48 (approved):

PR #482, opened 2026-04-10
Make search, assault and cqb tasks more resilient to locality changes

diwako:

1. Make search, assault and cqb tasks look for locality change when running
2. Exit task and reroute to where leader is now local to

### DIFF
--- a/addons/wp/functions/fnc_taskAssault.sqf
+++ b/addons/wp/functions/fnc_taskAssault.sqf
@@ -119,6 +119,15 @@ _group setFormation "LINE";
             {
                 params ["_args", "_handle"];
                 _args params ["_unit", "_group", "_retreat", "_threshold"];
+                if !(local _unit) exitWith {
+                    _handle call CBA_fnc_removePerFrameHandler;
+                    _unit setVariable [QGVAR(taskAssault), nil];
+
+                    // group
+                    private _groupMembers = _group getVariable [QGVAR(taskAssaultMembers), []];
+                    _group setVariable [QGVAR(taskAssaultMembers), _groupMembers - [_unit]];
+                };
+
                 private _destination = (_group getVariable [QGVAR(taskAssaultDestination), getPos _unit]) call CBA_fnc_getPos;
 
                 // exit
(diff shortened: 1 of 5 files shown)


addons/wp/functions/fnc_taskCreep.sqf
```diff
@@ -114,6 +114,11 @@ waitUntil {
     // performance
     waitUntil {sleep 1; simulationEnabled leader _group};
 
+    if !(local leader _group) exitWith {
+        [QGVAR(taskCreep), _this, leader _group] call CBA_fnc_targetEvent;
```
jokoho48, line 118: whe need to remove the eventhandlers here. and also remove them when they change locality.

addons/wp/functions/fnc_taskRush.sqf
```diff
@@ -127,6 +127,11 @@ waitUntil {
     // performance
     waitUntil { sleep 1; simulationEnabled (leader _group); };
 
+    if !(local leader _group) exitWith {
+        [QGVAR(taskRush), _this, leader _group] call CBA_fnc_targetEvent;
```
jokoho48, line 131: also here we need to remove the eventhandlers.

jokoho48 (changes requested):
with the eventhandler we also need to think about when the group is set as center position it self and it moves it may shifts the center when it gets recalled

addons/wp/functions/fnc_taskCreep.sqf
diwako: Not really. The event handlers will clean themselves up after they fire.
enableAttack and setCombatMode are bot LE and thus have no effect on remote units.

addons/wp/functions/fnc_taskAssault.sqf
```diff
@@ -212,6 +226,11 @@ waitUntil {
 // clean up
 _group setVariable [QGVAR(taskAssaultDestination), nil];
 _group setVariable [QGVAR(taskAssaultMembers), nil];
+
+if !(local leader _group) exitWith {
```
diwako, line 230: Just opening a conversation for this

> with the eventhandler we also need to think about when the group is set as center position it self and it moves it may shifts the center when it gets recalled

In my opinion that is fine, the main goal is to have some support for running tasks to survive the locality change.

PR #486, opened 2026-04-22
Per-Side Dynamic Reinforcement 

UnRealxInferno:
- Adds CBA options to enable dynamic reinforcement: Supports sandbox missions such as ALiVE via running when units spawn in, rather than at mission start (This may have a small performance impact in missions with large amounts of AI). This system is OFF by default.
- Adds CBA option to select maximum amount of groups called upon by the reinforcement system (0 by default, retaining original behaviour)

### DIFF
--- a/addons/danger/XEH_postInit.sqf
+++ b/addons/danger/XEH_postInit.sqf
@@ -19,3 +19,16 @@
     [_this select 0, QGVAR(OnReinforce), _this] call BIS_fnc_callScriptedEventHandler;
     [_this select 1, QGVAR(OnReinforce), _this] call BIS_fnc_callScriptedEventHandler;
 }] call CBA_fnc_addEventHandler;
+
+if (isServer) then {
+    {
+        [_x] call FUNC(setDynamicReinforcement);
+    } forEach allUnits;
+
+    addMissionEventHandler ["EntityCreated", {
+        params [["_entity", objNull, [objNull]]];
+        if (_entity isKindOf "CAManBase") then {
+            [_entity] call FUNC(setDynamicReinforcement);
+        };
+    }];
+};
--- a/addons/danger/XEH_preInit.sqf
+++ b/addons/danger/XEH_preInit.sqf
@@ -62,8 +62,9 @@ if (isNil QGVAR(dangerUntil)) then {
         private _adjustPos = _pos findEmptyPosition [5, 35, "Land_BagBunker_Large_F"];
         if (_adjustPos isNotEqualTo []) then {_pos = _adjustPos;};
 
+        private _fnc = FUNC(tacticsReinforce);
         {
-            [leader _x, _pos] call FUNC(tacticsReinforce);
+            [leader _x, _pos] spawn _fnc;
         } forEach _reinforceGroups;
     };
 
--- a/addons/danger/XEH_preInit.sqf
+++ b/addons/danger/XEH_preInit.sqf
@@ -32,7 +32,8 @@ if (isNil QGVAR(dangerUntil)) then {
 // EH handling reinforcement and combat mode
 [QEGVAR(main,OnInformationShared), {
     params [["_unit", objNull], "", ["_target", objNull], ["_groups", []]];
-    private _sideGroups = _groups select {(side _x) isEqualTo (side _unit)};
+    private _sideUnit = side _unit;
+    private _sideGroups = _groups select {(side _x) isEqualTo _sideUnit};
 
     // reinforce
     if (!isNull _target) then {
--- a/addons/danger/XEH_preInit.sqf
+++ b/addons/danger/XEH_preInit.sqf
@@ -32,28 +32,44 @@ if (isNil QGVAR(dangerUntil)) then {
 // EH handling reinforcement and combat mode
 [QEGVAR(main,OnInformationShared), {
     params [["_unit", objNull], "", ["_target", objNull], ["_groups", []]];
+    private _sideGroups = _groups select {(side _x) isEqualTo (side _unit)};
+
+    // reinforce
+    if (!isNull _target) then {
+        private _reinforceGroups = _sideGroups select {
+            private _leader = leader _x;
+            local _leader
+            && {_x getVariable [QGVAR(enableGroupReinforce), false]}
+            && {(_x getVariable [QGVAR(enableGroupReinforceTime), -1]) < time}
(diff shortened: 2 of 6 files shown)


addons/danger/XEH_postInit.sqf
```diff
@@ -19,3 +19,16 @@
     [_this select 0, QGVAR(OnReinforce), _this] call BIS_fnc_callScriptedEventHandler;
     [_this select 1, QGVAR(OnReinforce), _this] call BIS_fnc_callScriptedEventHandler;
 }] call CBA_fnc_addEventHandler;
+
+if (isServer) then {
+    {
+        [_x] call FUNC(setDynamicReinforcement);
```
rautamiekka, line 25: Since this is run by the server on bootup, it might be useful to `spawn` the func instead of `call`ing, possibly speeding up the process (ain't like the reinforcements are needed so early anyway, so nothing should be affected if the server will be processing this over time).

Caching the func for so many executions would definitely speed it up:
```suggestion
    private _fnc = FUNC(setDynamicReinforcement);

    {
        [_x] spawn _fnc;
```

addons/danger/XEH_preInit.sqf
```diff
@@ -32,28 +32,44 @@ if (isNil QGVAR(dangerUntil)) then {
...
+
+        // check for zero pos
+        if (_pos isEqualTo [0, 0, 0]) then {_pos = getPosASL _unit;};
+
+        // find free space
+        private _adjustPos = _pos findEmptyPosition [5, 35, "Land_BagBunker_Large_F"];
+        if (_adjustPos isNotEqualTo []) then {_pos = _adjustPos;};
+
+        {
+            [leader _x, _pos] call FUNC(tacticsReinforce);
```
rautamiekka: A similar thing as https://github.com/nk3nny/LambsDanger/pull/486/changes#r3125357069, except the reinforcements will have a bit of a delay in their acting and allowing groups to act simultaneously:
```suggestion
        private _fnc = FUNC(tacticsReinforce);
        {
            [leader _x, _pos] spawn _fnc;
```

addons/danger/XEH_preInit.sqf
```diff
@@ -32,28 +32,44 @@ if (isNil QGVAR(dangerUntil)) then {
 // EH handling reinforcement and combat mode
 [QEGVAR(main,OnInformationShared), {
     params [["_unit", objNull], "", ["_target", objNull], ["_groups", []]];
+    private _sideGroups = _groups select {(side _x) isEqualTo (side _unit)};
```
rautamiekka:
```suggestion
    private _sideUnit = side _unit;
    private _sideGroups = _groups select {(side _x) isEqualTo _sideUnit};
```

addons/danger/XEH_preInit.sqf
```diff
@@ -32,28 +32,44 @@ if (isNil QGVAR(dangerUntil)) then {
...
+
+    // reinforce
+    if (!isNull _target) then {
+        private _reinforceGroups = _sideGroups select {
+            private _leader = leader _x;
+            local _leader
```
rautamiekka, line 42: Presumably this code is still WIP, since the leader ain't referenced after this ?

addons/danger/XEH_postInit.sqf
jokoho48: postInit is schedueled anyway. no need to use spawn here. in theory wrapping it in a direct call would be even better.

addons/danger/XEH_preInit.sqf
UnRealxInferno: I think this was leftover from another iteration, I think it can be removed

addons/danger/XEH_postInit.sqf
```diff
@@ -19,3 +19,16 @@
...
+    } forEach allUnits;
```
jokoho48, line 26: why are we looping over all the Units here? reinforcement is only for groups. so we should only loop over groups. this does duplicated work not needed

addons/danger/XEH_postInit.sqf
```diff
@@ -19,3 +19,16 @@
...
+
+    addMissionEventHandler ["EntityCreated", {
```
jokoho48, line 28: this should probably be GroupCreated and not EntityCreated. no need to set that on every entity created

UnRealxInferno:
Added the suggestions you chaps made, I'm not the best with SQF but the system as-is does work and is tested.

addons/danger/XEH_preInit.sqf
```diff
@@ -32,28 +32,46 @@ if (isNil QGVAR(dangerUntil)) then {
...
+        // check for zero pos
+        if (_pos isEqualTo [0, 0, 0]) then {_pos = getPosASL _unit;};
+
+        // find free space
+        private _adjustPos = _pos findEmptyPosition [5, 35, "Land_BagBunker_Large_F"];
+        if (_adjustPos isNotEqualTo []) then {_pos = _adjustPos;};
+
+        private _fnc = FUNC(tacticsReinforce);
+        {
+            [leader _x, _pos] spawn _fnc;
```
jokoho48, line 68: we dont use spawn in lambs also why assign the function to a variable before using it?

addons/danger/XEH_preInit.sqf
jokoho48: we don't use spawn in Lambs. also spawn does now allow things to act simultaneously.

addons/danger/XEH_preInit.sqf
```diff
@@ -32,28 +32,46 @@ if (isNil QGVAR(dangerUntil)) then {
...
+        } forEach _reinforceGroups;
+    };
+
```
jokoho48, line 71: most of those changes don't make sense. and are overcomplicated. this can and should probably done way easier. also this is is now looping over the groups very often.

jokoho48 (changes requested):

addons/danger/XEH_preInit.sqf
rautamiekka: Caching a func (or more correctly: the reference to it) locally does the same it does with other vars: a lot faster subsequent access, since the engine doesn't have to leave contexts all the way to the global space to find the address before it can execute the func.

UnRealxInferno:
Gonna rebase into 2 PRs

PR #487, opened 2026-04-22
Per-Side Dynamic Reinforcements

UnRealxInferno:
Will add per-side dynamic reinforcements via an EH that runs on groupspawn.

### DIFF
--- a/addons/danger/XEH_postInit.sqf
+++ b/addons/danger/XEH_postInit.sqf
@@ -21,12 +21,13 @@
 }] call CBA_fnc_addEventHandler;
 
 addMissionEventHandler ["GroupCreated", {
+    params ["_group"]; 
     [{
-        [_this select 0] call FUNC(applySideReinforcementSetting);
-    }, [_this], 1] call CBA_fnc_waitAndExecute;
-}];
+        _this call FUNC(applySideReinforcementSetting);
+    }, _group, 1] call CBA_fnc_waitAndExecute;
+}];.
 
 {
-    [_x] call FUNC(applySideReinforcementSetting);
+    _x call FUNC(applySideReinforcementSetting);
 } forEach allGroups;
 
--- a/addons/danger/XEH_postInit.sqf
+++ b/addons/danger/XEH_postInit.sqf
@@ -20,22 +20,24 @@
     [_this select 1, QGVAR(OnReinforce), _this] call BIS_fnc_callScriptedEventHandler;
 }] call CBA_fnc_addEventHandler;
 
-GVAR(applySideReinforcementSetting) = {
-    params ["_group"];
-    if (isNull _group) exitWith {};
-
-    private _enabled = switch (side _group) do {
-        case west: {GVAR(enableReinforceWest)};
-        case east: {GVAR(enableReinforceEast)};
-        case independent: {GVAR(enableReinforceIndependent)};
-        default {false};
-    };
-
-    _group setVariable [QGVAR(enableGroupReinforce), _enabled, true];
-};
-
 addMissionEventHandler ["GroupCreated", {
     params ["_group"];
-    [_group] call GVAR(applySideReinforcementSetting);
+
+    [
+        {
+            params ["_group"];
+            !isNull _group && {local _group}
+        },
+        {
+            params ["_group"];
+            [_group] call FUNC(applySideReinforcementSetting);
+        },
+        [_group],
+        5
+    ] call CBA_fnc_waitUntilAndExecute;
 }];
+
+{
+    [_x] call FUNC(applySideReinforcementSetting);
+} forEach allGroups;
 
--- a/addons/danger/XEH_postInit.sqf
+++ b/addons/danger/XEH_postInit.sqf
@@ -20,22 +20,12 @@
     [_this select 1, QGVAR(OnReinforce), _this] call BIS_fnc_callScriptedEventHandler;
 }] call CBA_fnc_addEventHandler;
 
-GVAR(applySideReinforcementSetting) = {
-    params ["_group"];
-    if (isNull _group) exitWith {};
-
-    private _enabled = switch (side _group) do {
-        case west: {GVAR(enableReinforceWest)};
-        case east: {GVAR(enableReinforceEast)};
-        case independent: {GVAR(enableReinforceIndependent)};
-        default {false};
-    };
-
-    _group setVariable [QGVAR(enableGroupReinforce), _enabled, true];
-};
-
 addMissionEventHandler ["GroupCreated", {
     params ["_group"];
-    [_group] call GVAR(applySideReinforcementSetting);
+    [_group] call FUNC(applySideReinforcementSetting);
 }];
+
+{
+    [_x] call FUNC(applySideReinforcementSetting);
+} forEach allGroups;
 
--- a/addons/danger/XEH_postInit.sqf
+++ b/addons/danger/XEH_postInit.sqf
@@ -25,7 +25,7 @@ addMissionEventHandler ["GroupCreated", {
     [{
         _this call FUNC(applySideReinforcementSetting);
     }, _group, 1] call CBA_fnc_waitAndExecute;
-}];.
+}];
 
 {
     _x call FUNC(applySideReinforcementSetting);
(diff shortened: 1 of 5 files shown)


addons/danger/XEH_postInit.sqf
```diff
@@ -19,3 +19,23 @@
     [_this select 0, QGVAR(OnReinforce), _this] call BIS_fnc_callScriptedEventHandler;
     [_this select 1, QGVAR(OnReinforce), _this] call BIS_fnc_callScriptedEventHandler;
 }] call CBA_fnc_addEventHandler;
+
+GVAR(applySideReinforcementSetting) = {
```
jokoho48: this should be a own function and file.

addons/danger/XEH_postInit.sqf
```diff
@@ -19,3 +19,23 @@
...
+    if (isNull _group) exitWith {};
+
+    private _enabled = switch (side _group) do {
+        case west: {GVAR(enableReinforceWest)};
+        case east: {GVAR(enableReinforceEast)};
+        case independent: {GVAR(enableReinforceIndependent)};
+        default {false};
+    };
+
+    _group setVariable [QGVAR(enableGroupReinforce), _enabled, true];
```
jokoho48: i think it would be better if you instead of using _enabled here check if it is enabled and if not then just skip it. that way you dont overwrite possible mission scripts

addons/danger/XEH_postInit.sqf
```diff
@@ -19,3 +19,23 @@
     [_this select 0, QGVAR(OnReinforce), _this] call BIS_fnc_callScriptedEventHandler;
     [_this select 1, QGVAR(OnReinforce), _this] call BIS_fnc_callScriptedEventHandler;
 }] call CBA_fnc_addEventHandler;
+
+GVAR(applySideReinforcementSetting) = {
+    params ["_group"];
+    if (isNull _group) exitWith {};
```
jokoho48: this should probably not just check for isNull but also for local, we dont need/want to spam network with the resetting of that value

jokoho48 (changes requested):
the startup loop that goes over all groups that are created in the editor/mission is missing here.

addons/danger/functions/fnc_applySideReinforcementSetting.sqf
jokoho48, line 1: i would probably look at how you did it in the old [PR](https://github.com/nk3nny/LambsDanger/pull/486/files#diff-4357640182c8e475e80550a9d67d1b41b249f3b45b0c804221cf0b65b88d2e5f). this now has quite a bit of code duplication in. also at best set a accurate author in the header.(what should in theory be you/your username)

jokoho48 (changes requested):

addons/danger/XEH_postInit.sqf
```diff
@@ -19,3 +19,13 @@
     [_this select 0, QGVAR(OnReinforce), _this] call BIS_fnc_callScriptedEventHandler;
     [_this select 1, QGVAR(OnReinforce), _this] call BIS_fnc_callScriptedEventHandler;
 }] call CBA_fnc_addEventHandler;
+
+addMissionEventHandler ["GroupCreated", {
+    params ["_group"];
+    [_group] call FUNC(applySideReinforcementSetting);
```
jokoho48: we may need to think about adding a CBA_fnc_waitUntilAndExecute here, because i am not 100% sure if all the group informations are already finished and setup here. especially in a Multiplayer setting.

addons/danger/functions/fnc_applySideReinforcementSetting.sqf
UnRealxInferno: Yeah the author bit had been autocompleted for some bizzare reason

addons/danger/XEH_postInit.sqf
UnRealxInferno: I've looked at this and not seen an issue in MP yet but will add

jokoho48: sorry i meant wait and execute. just waiting for like 1 second.

addons/danger/XEH_postInit.sqf
```diff
@@ -19,3 +19,17 @@
...
     [_this select 1, QGVAR(OnReinforce), _this] call BIS_fnc_callScriptedEventHandler;
 }] call CBA_fnc_addEventHandler;
+
+addMissionEventHandler ["GroupCreated", {
+    params ["_group"];
+
+    [{
+        params ["_group"];
+        [_group] call FUNC(applySideReinforcementSetting);
+    }, [_group], 0] call CBA_fnc_waitAndExecute;
```
jokoho48: 0 seconds makes no sense. this should be 1 second also you dont need to add a [] around everything. nor do you need params here. in theory _this should be totally fine

jokoho48 (changes requested):

addons/danger/XEH_postInit.sqf
```diff
@@ -19,3 +19,14 @@
...
+
+addMissionEventHandler ["GroupCreated", {
+    [{
+        [_this select 0] call FUNC(applySideReinforcementSetting);
+    }, [_this], 1] call CBA_fnc_waitAndExecute;
+}];
+
+{
+    [_x] call FUNC(applySideReinforcementSetting);
+} forEach allGroups;
```
jokoho48, line 32: i mean it like that. if you only want to pass 1 piece of data to a function you dont need to create an array for that. just pass the data.

```suggestion
addMissionEventHandler ["GroupCreated", {
    params ["_group"]; 
    [{
        _this call FUNC(applySideReinforcementSetting);
    }, _group, 1] call CBA_fnc_waitAndExecute;
}];.

{
    _x call FUNC(applySideReinforcementSetting);
} forEach allGroups;
```

jokoho48 (approved):

addons/danger/functions/fnc_applySideReinforcementSetting.sqf
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [group player] call FUNC(applySideReinforcementSetting)
+ *
+ * Public: No
+ */
+params ["_group", grpNull, [grpNull]];
```
jokoho48: this needs to be 
```suggestion
params [["_group", grpNull, [grpNull]]];
```

addons/danger/XEH_postInit.sqf
```diff
@@ -19,3 +19,15 @@
     [_this select 0, QGVAR(OnReinforce), _this] call BIS_fnc_callScriptedEventHandler;
     [_this select 1, QGVAR(OnReinforce), _this] call BIS_fnc_callScriptedEventHandler;
 }] call CBA_fnc_addEventHandler;
+
+addMissionEventHandler ["GroupCreated", {
+    params ["_group"]; 
+    [{
+        _this call FUNC(applySideReinforcementSetting);
+    }, _group, 1] call CBA_fnc_waitAndExecute;
+}];.
```
jokoho48: it seems a . have snuck in to this.
```suggestion
}];
```

addons/danger/stringtable.xml
```diff
@@ -215,6 +215,25 @@
             <German>Einheiten bei denen diese Option aktiviert ist haben eine erhöhte Kommunikationsreichweite.\n Dieser Effekt kann auch erzeugt werden durch das Ausrüsten der Einheit mit einem standard Radio-Rucksack oder einem TFAR-Mod kompatiblen Radio.</German>
             <Chinesesimp>共享信息时切换了无线电的装置提高了通信范围\n这种效果也可以通过为装置配备普通无线电背包或TFAR mod启用的无线电来实现</Chinesesimp>
         </Key>
+
```
jokoho48: this new line needs to be deleted. that should fix the issue with the stringtable linter

jokoho48 (changes requested):

addons/danger/XEH_postInit.sqf
UnRealxInferno: typo from me goddamnit

jokoho48 (approved):

UnRealxInferno:
thank god

PR #489, opened 2026-05-11
Garrison - Allow a specific set of exit conditions.

Alien314:

1.  Update function description.
2. Allow an array to be passed to _exitCondition to activate a set of exit conditions other than "All".
3. Allow random to select the suppressed event exitCondition.

### DIFF
--- a/addons/wp/functions/fnc_taskGarrison.sqf
+++ b/addons/wp/functions/fnc_taskGarrison.sqf
@@ -203,6 +203,12 @@ private _fnc_addEventHandler = {
     // set EH
     _unit setVariable [QGVAR(eventhandlers), _ehs];
 };
+// determine exit condition(s) to use
+if !(_exitCondition isEqualType []) then { 
+    if (_exitCondition == -1) then { _exitCondition = [1,2,3,4]; } else { _exitCondition = [_exitCondition]; };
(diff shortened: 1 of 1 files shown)


addons/wp/functions/fnc_taskGarrison.sqf
```diff
@@ -241,14 +241,12 @@ private _fnc_addEventHandler = {
...
     };
 
-    if (_exitCondition == -1) then {
-        for "_i" from 0 to 4 do {
-            [_x, _i] call _fnc_addEventHandler;
-        };
-    } else {
-        [_x, _exitCondition] call _fnc_addEventHandler;
+    if (_exitCondition == -1) then { _exitCondition = [1,2,3,4]; };
+    if !(_exitCondition isEqualType []) then { _exitCondition = [_exitCondition]; };
```
diwako: These two lines of evaluation can be done above the `forEach` over each unit in the squad so it does not need to evaluate for every unit.

Then you can just do this just without the IF in the end for it.
```sqf
private _unit = _x;
{ [_unit, _x] call _fnc_addEventHandler; } forEach _exitCondition;
```

Additionally, use `isEqualTo` when comparing to -1 with a potential array

